PR #3, opened 2019-06-16
Move TxCoinAgePriority to its sole usage location.

schancel:
Summary:
TxCoinAgePriority is currently only used by `addPriorityTxns`.  Therefore, moving it closer will help ensure that it is removed when `addPriorityTxns` is removed.

Test Plan:
  make check && ./test/functional/test_runner.py

Tags: #cleanup


schancel:
This isn't worth having open anymore. Can redo later.

PR #11, opened 2021-02-22
ASERT from genesis

EyeOfPython:
- Block#1 as ASERT anchor block (uses timestamp of genesis).
- Use `GetNextEDAWorkRequired` for the genesis block and `GetNextASERTWorkRequired` for every other block.
- Remove `asert_activation_anchor_test` test, because there's no activation of ASERT anymore.
- Fix `CreateNewBlock_validity`; set proper `nBits` for test blocks using `GetNextWorkRequired` and use new externally mined nonces.
- Temporarily disable `test/functional/p2p_dos_header_tree.py`, as this relies on blocks & checkpoints from a testnet (with ASERT), which currently does not exist yet.


src/pow/aserti32d.cpp
```diff
@@ -22,69 +22,13 @@ const CBlockIndex *GetASERTAnchorBlockCache() noexcept {
...
-    }
-
-    // Overwrite the cache with the anchor we found. More likely than not, the
-    // next time we are asked to validate a header it will be part of same /
-    // similar chain, not some other unrelated chain with a totally different
-    // anchor.
-    cachedAnchor = anchor;
-
-    return anchor;
+    return pindex->GetAncestor(1);
```
schancel, line 31: Beautiful

src/pow/pow.cpp
```diff
@@ -31,14 +31,10 @@ uint32_t GetNextWorkRequired(const CBlockIndex *pindexPrev,
...
-    if (IsAxionEnabled(params, pindexPrev)) {
+    if (pindexPrev->pprev) {
         return GetNextASERTWorkRequired(pindexPrev, pblock, params);
     }
 
-    if (IsDAAEnabled(params, pindexPrev)) {
-        return GetNextDAAWorkRequired(pindexPrev, pblock, params);
-    }
-
     return GetNextEDAWorkRequired(pindexPrev, pblock, params);
```
schancel, line 38: This is used for the Genesis block and nothing else? Something seems fishy here.

src/pow/pow.cpp
```diff
@@ -31,14 +31,10 @@ uint32_t GetNextWorkRequired(const CBlockIndex *pindexPrev,
         return pindexPrev->nBits;
     }
 
-    if (IsAxionEnabled(params, pindexPrev)) {
+    if (pindexPrev->pprev) {
```
schancel, line 34: Hmmm

schancel: Can we flip this to:

```
if (!pindexPrev->pprev) {
    const uint32_t nProofOfWorkLimit =
        UintToArith256(params.powLimit).GetCompact();
    return nProofOfWorkLimit;
}

return GetNextASERTWorkRequired(pindexPrev, pblock, params);
```

schancel (changes requested):

src/pow/pow.cpp
EyeOfPython: `GetNextEDAWorkRequired` is just used for the block after genesis.

ASERT requires two existing blocks.

I decided it's easiest to keep the old code for this one block; we can add a special case to GetNextASERTWorkRequired later.

Or what do you have in mind?

src/pow/pow.cpp
EyeOfPython: Ok that answers the question.

src/pow/pow.cpp
```diff
@@ -31,15 +31,19 @@ uint32_t GetNextWorkRequired(const CBlockIndex *pindexPrev,
...
-    if (IsAxionEnabled(params, pindexPrev)) {
-        return GetNextASERTWorkRequired(pindexPrev, pblock, params);
+    // ASERT requires an anchor block which has one parent.
+    // The Genesis block has no parent. Therefore, the block after Genesis
+    // will be our anchor block.
+    // We assign it minimum difficulty.
+    if (!pindexPrev->pprev) {
+        const uint32_t nProofOfWorkLimit =
+            UintToArith256(params.powLimit).GetCompact();
+        return nProofOfWorkLimit;
```
schancel, line 41: Can you add a TODO or a task in this repo to delete the EDA code?

src/pow/aserti32d.cpp
schancel: Can we leave this in place (spitballing):

```
    const CBlockIndex *lastCached = cachedAnchor.load();
    if (lastCached && pindex->GetAncestor(lastCached->nHeight) == lastCached) {
        return lastCached;
    }

    const CBlockIndex *anchor = pindex->GetAncestor(1);
    
    // Overwrite the cache with the anchor we found. More likely than not, the
    // next time we are asked to validate a header it will be part of same /
    // similar chain, not some other unrelated chain with a totally different
    // anchor.
    cachedAnchor = anchor;

    return anchor;
```

Then we can just remove the `pindex->GetAncestor(lastCached->nHeight) == lastCached` check once we set a checkpoint after that value.

schancel (changes requested):

schancel (approved):

src/pow/pow.cpp
EyeOfPython: In fact, the EDA, DAA and Grasberg code. Grasberg makes 0 sense for our coin.

schancel (approved):

schancel:
Looks good to mel

PR #12, opened 2021-02-22
Coinbase rule from genesis

EyeOfPython:
- Enforce whitelist from genesis in `GetMinerFundWhitelist`.
- Set `minerFundMinValue` to miner fund value from genesis in `getblocktemplate`.
- Update `CreateNewBlock_validity`; keep additional coinbase outputs generated by `getblocktemplate` and add `MINERREWARD` calculated from `GetMinerFundAmount`.
- Remove activation logic from `test/functional/abc_feature_minerfund.py`
- Remove activation logic from `test/functional/abc_mining_basic.py`, and verify from genesis.

NOTE: in `test/functional/abc_mining_basic.py`, `getblocktemplate` always fails if called from genesis, where it reports: `JSONRPC error: Bitcoin ABC is in initial sync and waiting for blocks...`.

This needs to be fixed before mining operations can start.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -28,6 +28,8 @@
 #include <boost/test/unit_test.hpp>
 
 #include <memory>
+
+#include <minerfund.h>
 #include <pow/pow.h>
 
 namespace miner_tests {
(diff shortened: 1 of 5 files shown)


test/functional/abc_mining_basic.py
```diff
@@ -49,88 +48,42 @@ def assert_getblocktemplate(expected):
...
-                },
-            },
-            # Although the coinbase value need not necessarily be the same as
-            # the last block due to halvings and fees, we know this to be true
-            # since we are not crossing a halving boundary and there are no
-            # transactions in the mempool.
-            'coinbasevalue': block_reward * COIN,
-            'mintime': AXION_ACTIVATION_TIME + 1,
-        })
+        # TODO: Cannot use getblocktemplate for block after genesis
```
schancel, line 56: Is this possibly caused by this line? 

```
    if (m_chain.Tip()->nChainWork < nMinimumChainWork) {
        return true;
    }
```

I *really* dislike `::IsInitialBlockDownload()` it caused me some other problems in the past. From my perspective this function is a heuristic mess, and there should be some clear chainstate management at someday.... But not today.

src/test/miner_tests.cpp
```diff
@@ -58,36 +60,43 @@ constexpr static struct {
...
-    {4, 0xdee3b283}, {4,  0x9e0eef78}, {4, 0x0f87df2f}, {4,  0x8b76a010},
-    {2, 0xefc5c3e4}, {2,  0x32662ae3}, {2, 0xa8f47138}, {2,  0xc550b341},
-    {3, 0xa550cb97}, {2,  0x55955315}, {2, 0x4c3f3038}, {4,  0x1658a340},
-    {7, 0x29f9333b}, {3,  0xdf97ae03}, {5, 0x321aa10b}, {2,  0x9f3c21cf},
-    {3, 0x3568a31c}, {4,  0xceef8301}, {3, 0x09645c63}, {4,  0xb328d56f},
-    {4, 0xc4de303b}, {2,  0x50f80a5b}, {3, 0x1455e14c}, {4,  0x3da25948},
-    {2, 0x1fbb494e}, {8,  0xa41ad1b4}, {2, 0xee911aa6}, {2,  0xaea15a0c},
-    {8, 0xbcc24752}, {3,  0x41aa8978}, {2, 0x41795716}, {6,  0x7d6e019},
-    {3, 0xeb30b104}, {2,  0x88ba4d3c},
+    {1, 0xa9d9c5fe}, {2, 0xc4b8f377}, {3, 0x8cec161a}, {2, 0xc8925628},
```
schancel, line 63: Joy. It seems like this will keep mutating. Is there a way to avoid this?

schancel (approved):

src/test/miner_tests.cpp
EyeOfPython: No. Other than frying my machine, it's not a problem though, is it? Keeps the commits consistent.

test/functional/abc_mining_basic.py
EyeOfPython: Add it to the TODO list, because this actually needs to be fixed before we get on mainnet.

src/test/miner_tests.cpp
schancel: I'm just trying to understand the tests. Bitcoind is ~300K LOC, and I have only messed with fees, mempool, and mining. I have ~0 experience with validation code.

schancel (approved):

PR #13, opened 2021-02-22
Enforce Sigchecks from genesis

EyeOfPython:
- Remove `SCRIPT_ENFORCE_SIGCHECKS` flag.
- Don't use `TxSigCheckLimiter::getDisabled`.
- Keep `SCRIPT_VERIFY_INPUT_SIGCHECKS`, as many tests (esp. script_tests) rely on many sig checks being allowed (e.g. 20 out of 20 multisig).


schancel (approved):
Tasty. Need to do some kind of rebase or something though with these PRs. Not sure why yours is so divergent from master after I updated it.

PR #21, opened 2021-02-25
Cleanstack along p2sh

EyeOfPython:
- Move `SCRIPT_VERIFY_CLEANSTACK` into `MANDATORY_SCRIPT_VERIFY_FLAGS` from `STANDARD_SCRIPT_VERIFY_FLAGS`.
- Enforce `SCRIPT_VERIFY_CLEANSTACK` along with `SCRIPT_VERIFY_P2SH`. Eventually, both can be removed.
- This does not remove `SCRIPT_VERIFY_CLEANSTACK` (despite it being mandatory now) because too many tests depend on allowing stack items being left on the stack.


src/validation.cpp
```diff
@@ -1524,6 +1524,7 @@ static uint32_t GetNextBlockScriptFlags(const Consensus::Params &params,
     // Start enforcing P2SH (BIP16)
     if ((pindex->nHeight + 1) >= params.BIP16Height) {
         flags |= SCRIPT_VERIFY_P2SH;
+        flags |= SCRIPT_VERIFY_CLEANSTACK;
```
schancel, line 1527: Why start this here?

schancel (approved):

schancel:
Approved, but the BIP16 height stuff will need to be set to 0 eventually and removed.

PR #22, opened 2021-02-25
Ctor from genesis

EyeOfPython:
- Remove logic causing CTOR to activate and always keep it as active.
- Fix `CreateNewBlock_validity`, which relied on TTOR.


schancel (approved):

schancel:
Looks good, but extra commits.

PR #24, opened 2021-02-26
Strict sigs from genesis

EyeOfPython:
- Remove `SCRIPT_VERIFY_STRICTENC`.
- Remove `SCRIPT_VERIFY_DERSIG`.
- Remove `SCRIPT_VERIFY_LOW_S`.

Note: I chose to remove all script flags at once, as they are not really independent. For example, requiring LOW_S for all txs also implies STRICTENC in some way. Therefore, just packaging them in one large commit seemed easier to reason about.

- `OutputTx` now requires a secp256k1 instance (to display sigs), so `Secp256k1Init` creation is moved into `CommandLineRawTx`.
- `script/sigencoding.cpp`: Now always strictly enforces signature encoding .
- `test/checkdatasig_tests.cpp`: Now no longer checks different things for different encoding flags; instead always checks strictly.
- `test/data/script_tests.json`:
  - Remove `STRICTENC`, `DERSIG` and `LOW_S` flags.
  - Remove test variants that check if an incorrectly encoded signature succeeds if flags are missing.
- `test/data/tx_valid.json`:
  - Normalize all signatures.
  - Move now invalid txs into `tx_invalid.json`.
- Util tests: Fix all sigs


schancel:
Looks good assuming the hex values are sensible. I don't know what mutated?

schancel (approved):

PR #26, opened 2021-02-27
BIP-68 (LOCKTIME_VERIFY_SEQUENCE) from genesis

EyeOfPython:
- Remove `LOCKTIME_VERIFY_SEQUENCE`.
- Mitigate buggy miner_tests test by setting nSequence to valid values before adding them to the mempool.

### DIFF
--- a/src/consensus/tx_verify.cpp
+++ b/src/consensus/tx_verify.cpp
@@ -83,8 +83,7 @@ std::pair<int, int64_t> CalculateSequenceLocks(const CTransaction &tx,
     // tx.nVersion is signed integer so requires cast to unsigned otherwise
     // we would be doing a signed comparison and half the range of nVersion
     // wouldn't support BIP 68.
-    bool fEnforceBIP68 = static_cast<uint32_t>(tx.nVersion) >= 2 &&
-                         flags & LOCKTIME_VERIFY_SEQUENCE;
+    bool fEnforceBIP68 = static_cast<uint32_t>(tx.nVersion) >= 2;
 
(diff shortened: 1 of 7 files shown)


src/bitcoin-tx.cpp
```diff
@@ -847,6 +842,9 @@ static int CommandLineRawTx(int argc, char *argv[],
             startArg = 1;
         }
 
+        std::unique_ptr<Secp256k1Init> ecc;
```
schancel, line 845: I don't understand what is changing here. It looks behaviorally different?

src/consensus/tx_verify.cpp
```diff
@@ -83,8 +83,7 @@ std::pair<int, int64_t> CalculateSequenceLocks(const CTransaction &tx,
     // tx.nVersion is signed integer so requires cast to unsigned otherwise
     // we would be doing a signed comparison and half the range of nVersion
     // wouldn't support BIP 68.
-    bool fEnforceBIP68 = static_cast<uint32_t>(tx.nVersion) >= 2 &&
-                         flags & LOCKTIME_VERIFY_SEQUENCE;
+    bool fEnforceBIP68 = static_cast<uint32_t>(tx.nVersion) >= 2;
```
schancel, line 86: Should we set this == 2? There aren't anything else right now, and we probably want to restrict it in the future. Though, I think there is a check for nVersion == 2 now.

On BCH this was sort of required because there were old transactions with the version set oddly.

schancel:
Last commit seems okay.

schancel (approved):

PR #27, opened 2021-03-02
Mtp locktime from genesis

EyeOfPython:
- Remove `LOCKTIME_MEDIAN_TIME_PAST` flag.
- Always use GetMedianTimePast() for CLTV calculations.
- Fix `test/functional/feature_csv_activation.py`, `bip113signed1` and `bip113signed2` are now invalid.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -149,11 +149,7 @@ BlockAssembler::CreateNewBlock(const CScript &scriptPubKeyIn) {
     }
 
     pblock->nTime = GetAdjustedTime();
-    nMedianTimePast = pindexPrev->GetMedianTimePast();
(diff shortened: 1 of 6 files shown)


src/miner.cpp
```diff
@@ -149,11 +149,7 @@ BlockAssembler::CreateNewBlock(const CScript &scriptPubKeyIn) {
     }
 
     pblock->nTime = GetAdjustedTime();
-    nMedianTimePast = pindexPrev->GetMedianTimePast();
```
schancel, line 152: We're going to have to have a voice call for me to review this. I don't fully understand what changes semantically for locktime stuff.

schancel (approved):

PR #38, opened 2021-03-06
Disable nops

EyeOfPython:


schancel:
I'll look into this and see if I can't figure out the tests.

PR #43, opened 2021-03-07
Disable invalid opcodes (fail script even if not executed)

EyeOfPython:
- Disables the following opcodes:
  - `OP_RESERVED`
  - `OP_VER`
  - `OP_RESERVED1`
  - `OP_RESERVED2`
  - All opcodes after (and including) `FIRST_UNDEFINED_OP_VALUE`
- Fixes tests accordingly.


EyeOfPython:
This should be split into multiple smaller PRs.

PR #44, opened 2021-03-07
Enforce SIGHASH_FORKID from genesis

EyeOfPython:
- Makes sure `SCRIPT_ENABLE_SIGHASH_FORKID` is always enabled.
- This actually breaks no tests. Presumably, we fixed those in a previous commit.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1524,10 +1524,8 @@ static uint32_t GetNextBlockScriptFlags(const Consensus::Params &params,
     // Always enforce CLEANSTACK
     flags |= SCRIPT_VERIFY_CLEANSTACK;
 
-    // If the UAHF is enabled, we start accepting replay protected txns
-    if (IsUAHFenabled(params, pindex)) {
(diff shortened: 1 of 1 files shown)


schancel (changes requested):
Seems like there are two things happening here.

EyeOfPython:
This should fix it...

schancel (approved):

src/validation.cpp
```diff
@@ -1524,10 +1524,8 @@ static uint32_t GetNextBlockScriptFlags(const Consensus::Params &params,
     // Always enforce CLEANSTACK
     flags |= SCRIPT_VERIFY_CLEANSTACK;
 
-    // If the UAHF is enabled, we start accepting replay protected txns
-    if (IsUAHFenabled(params, pindex)) {
```
schancel, line 1528: Are we going to remove these functions as some point?

EyeOfPython: Yes. Make an issue

PR #62, opened 2021-03-08
Disable OP_SHA1

EyeOfPython:
SHA1 is now broken: https://shattered.io/

There's no reason to keep it. We can repurpose the opcode for some useful hashing algorithm, e.g. SHA3


src/script/interpreter.cpp
```diff
@@ -83,12 +83,18 @@ static void CleanupScriptCode(CScript &scriptCode,
 static bool IsOpcodeDisabled(opcodetype opcode, uint32_t flags) {
     switch (opcode) {
         case OP_RESERVED:
+        case OP_VER:
```
schancel, line 86: Are these things supposed to be in this commit?

schancel (changes requested):

src/script/interpreter.cpp
EyeOfPython: No. Will fix.

schancel:
Just one thing to mention, I would be careful to repurpose too many things later. disabling stuff that's borked is fine, but there's txn construction libraries to consider.  I forked dartsv for Cashew, but I would prefer not to have to change it much for now. It's very broken.

schancel (approved):

PR #70, opened 2021-03-09
Change blocktime to 2 minutes

EyeOfPython:
- Set `nPowTargetSpacing` to 120 seconds.
- Reduce `STALE_CHECK_INTERVAL` to 120 seconds.
- Fix ASERT tests; as corrolary they are now variable to `t_block`.
- Redo PoW in `miner_tests.cpp`.
- Reduce `GetBlockProofEquivalentTime` required for `assumevalid` to 5 days from 14 days (otherwise `feature_assumevalid.py` takes forever)
- Bump number of blocks in `feature_assumevalid.py` to satify `GetBlockProofEquivalentTime`.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -60,34 +60,34 @@ constexpr static struct {
     uint8_t extranonce;
     uint32_t nonce;
 } blockinfo[] = {
-    {1,  0x7458c333}, {2, 0x4f87692f}, {3,  0xe4448bee}, {3,  0xcd574612},
-    {3,  0x5bc18aa4}, {5, 0xdbaeec2c}, {3,  0x2667b5c0}, {7,  0x8601b02b},
-    {3,  0xfb9632cc}, {6, 0x3227743e}, {3,  0x5f8eb6ac}, {4,  0x995aade0},
-    {5,  0xf83707af}, {6, 0x5eaeacc8}, {5,  0x8904dd0f}, {5,  0x1777db09},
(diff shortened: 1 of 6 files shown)


EyeOfPython:
Things left to do:
- Write functional test checking for the new blocktime explicitly.

src/test/miner_tests.cpp
```diff
@@ -238,7 +238,7 @@ void MinerTestingSetup::TestPackageSelection(
     tx.vout[0].nValue = (100000000 - 10000) * SATOSHI;
     m_node.mempool->addUnchecked(entry.Fee(10000 * SATOSHI).FromTx(tx));
     pblocktemplate = AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey);
-    BOOST_CHECK_EQUAL(pblocktemplate->block.vtx[4]->GetId(), lowFeeTxId2);
+    BOOST_CHECK_EQUAL(pblocktemplate->block.vtx[1]->GetId(), lowFeeTxId2);
```
schancel, line 241: Why's the blocktime change the txn ordering?

schancel (changes requested):
Requesting changes until functional test added.

src/test/miner_tests.cpp
EyeOfPython: Will review

EyeOfPython: Found it: https://github.com/EyeOfPython/cashd/blob/blocktime_2_minutes/src/test/miner_tests.cpp#L561

Here, the nSequence gets set to the MTP. But the MTP changes because of the faster blocktime. So the hash changes.

Expected behavior. Well spotted!

schancel (approved):

PR #71, opened 2021-03-09
Increase script numbers to 63-bit + sign bit

EyeOfPython:
- Use `std::stoll` instead of `atoi64` for parsing int64_t (otherwise we can't detect out of bounds numbers).
- Increase `MAXIMUM_ELEMENT_SIZE` to 8.
- Use `__builtin_saddll_overflow` for additions, throw `scriptnum_error` if overflow.
- Use `__builtin_ssubll_overflow` for additions, throw `scriptnum_error` if overflow.
- Fix `script_tests.json`
- Fix BIN2NUM and OP_DIV/OP_MOD tests
- Remove `scriptnum10.h` and tests based on that, it's no longer useful. We don't need to test whether our numbers are compatible with Bitcoin Core 0.10.
- Fix `bitcoin-util-tests`.


src/core_read.cpp
```diff
@@ -74,16 +74,27 @@ CScript ParseScript(const std::string &s) {
...
+            int64_t n;
             // Number
-            int64_t n = atoi64(w);
+            try {
+                n = std::stoll(w);
+            } catch (...) {
+                is_valid = false;
+            }
+
+            if (n == std::numeric_limits<int64_t>::min()) {
```
schancel, line 86: If w doesn't parse, why do we want to do this check?

EyeOfPython: Cuz -0x8000000000000000 is a valid int64_t but an invalid script number

EyeOfPython:
Things left to do:
- Write CPP tests grilling the new numbers.
- Add more script_tests.json tests.
- Add a functional test just to make sure our new numbers don't do anything funky during block validation.

src/core_read.cpp
schancel: Yes. It just seems like if the parsing fails there's no reason to *also* do this check. Couldn't it be moved into the try block?

EyeOfPython: Can you propose a suggestion? My C++ is not good enough for that

EyeOfPython: I think I got it. This should be what you meant.

schancel (changes requested):
Going to mark request changes so it doesn't accidentally get merged.

EyeOfPython:
Posting Calin's feedback here:

> idk to be honest i tend to stay away from compiler extensions

> now granted bitcoin is pretty much set up to only compile on gcc and clang

> but to me it's always a bit of a questionable decision... although maybe in this case it's ok

> the only time it's worth it is in critical performance-sensitive code

> but yeah i mean this looks legit for what he's doing

> i will point out that inline here is superfluous but we won't focus on that...

> but yeah this looks like a good approach -- the builtins do work.

> the thing is though.. you shoudl put a static_assert in here somewhere

> static_assert(std::numeric_limits<long long int>::max() == std::numeric_limits<int64_t>::max());

> there is no guarantee that long long int and int64_t are the same size.. you want to really be sure of that. otherwise this code blows up.

> maybe put that in "assumptions.h" or whatever that file is that does this crap...

> honestlty i would wrap all those calls in some util function like:

> overflow_add overflow_sub, etc..

> and i would maybe use intrinsics on gcc

> but also write fallback code that doesn't use intrinsics

> there is somethign kind of obscene about seeing those compiler extensions just bare and naked in a header like that.

src/script/script.h
```diff
@@ -287,13 +291,19 @@ class CScriptNum {
     }
 
     inline CScriptNum operator/(const int64_t &rhs) const {
+        // Sanity check for invalid script num; should never occur
+        assert(m_value != std::numeric_limits<int64_t>::min());
+        assert(rhs != 0);
```
schancel, line 296: What's the resulting error for these compared to existing behavior?

src/script/script.h
```diff
@@ -414,6 +425,66 @@ class CScriptNum {
...
+
+    /**
+     * Computes a - b and stores it in result. If the difference overflows or
+     * results in an invalid Script integer, return true, otherwise false.
+     */
+    static bool overflow_sub_int64(int64_t a, int64_t b, int64_t &result) {
+#if HAVE_DECL___BUILTIN_SSUBLL_OVERFLOW
+        static_assert(std::numeric_limits<long long int>::max() ==
+                      std::numeric_limits<int64_t>::max());
+        long long int tmp_result;
```
schancel, line 466: Why this instead of `int64_t` ?

EyeOfPython: Otherwise we get a compile error. Both types are not treated the same.

src/script/script.h
EyeOfPython: I just added these asserts to make sure if there's any error in using these, the tests will blow up. Especially the -0x8000 0000 0000 0000 bug could be very subtle.

schancel (approved):

PR #78, opened 2021-03-10
Default to -acceptnonstdtxn=0 on testnet

### LINKED ISSUE #76: Enforce standardness on testnet
There’s no point in keeping non-standardness as default.

EyeOfPython:
Closes #76.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -175,7 +175,6 @@ public:
             pnSeed6_main, pnSeed6_main + ARRAYLEN(pnSeed6_main));
 
         fDefaultConsistencyChecks = false;
-        fRequireStandardPolicy = true;
         m_is_test_chain = false;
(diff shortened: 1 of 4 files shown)


src/chainparams.cpp
```diff
@@ -175,7 +175,6 @@ class CMainParams : public CChainParams {
             pnSeed6_main, pnSeed6_main + ARRAYLEN(pnSeed6_main));
 
         fDefaultConsistencyChecks = false;
-        fRequireStandardPolicy = true;
```
schancel, line 178: Are we deleting these because we are never going to allow testnet to be different for defaults in the future?

EyeOfPython: Yes. Why would we? Testnet and mainnet should be compatible.

schancel (approved):

PR #79, opened 2021-03-10
Relax OP_AND, OP_OR and OP_XOR.

### LINKED ISSUE #75: Relax OP_AND, OP_OR, OP_XOR
Bitwise operators like AND, OR and XOR are often used for arithmetic.

However, since OP_AND, OP_OR and OP_XOR in Script require the arguments to be of the same length and because numbers are minimally encoded in Script, they often can't be used in that context.

For example, say we want to set the least 4 significant bits of a number to 1, then we'd like to use the following script:

`<num> OP_15 OP_OR` (2 opcodes, 2 bytes)

However, this fails for all inputs that have more than 8 bits (or 7 if values are unsigned), since 0x0f has 1 byte.

Instead, one has to do this (assuming 63+1-bit integers):

`<num> OP_8 OP_NUM2BIN 0x000000000000000f OP_OR OP_BIN2NUM` (3 opcodes, 13 bytes),

or

`<num> OP_8 OP_NUM2BIN OP_15 OP_8 OP_NUM2BIN OP_OR OP_BIN2NUM` (4 opcodes, 7 bytes),

Which is very inefficient.

One possible fix is to allow arguments of different length, pick the longest of both arguments, padding the other one with zeros, and then applying the bitwise operator.

From time to time, it would still require a OP_BIN2NUM after the operator if superfluous zeros remain, but that would be required in the current workaround, too, so there are no additional costs.

EyeOfPython:
Closes #75.

- In `interpreter.cpp`, modifies the longer of both arrays instead of requiring both to be equal. Removes the shorter one from the stack.
- In `monolith_opcodes_tests.cpp`, in `RunTestForAllBitwiseOpcodesSizes`, also test all arrays longer than the outer loop. Also, remove failure test for unequal length.
- In script_tests.json, fix test for unequal length.


src/test/monolith_opcodes_tests.cpp
```diff
@@ -139,14 +139,34 @@ static void RunTestForAllBitwiseOpcodesSizes(const valtype &a, const valtype &b,
...
         wa.push_back(a[i]);
         wb.push_back(b[i]);
         wand.push_back(expected_and[i]);
         wor.push_back(expected_or[i]);
         wxor.push_back(expected_xor[i]);
-
+        // Test for len(wa) == len(wb)
         RunTestForAllBitwiseOpcodes(wa, wb, wand, wor, wxor);
+        // Test for all len(wb) == len(wa)..len(b), expand arrays
+        for (size_t j = wa.size(); j < wb.size(); j++) {
```
schancel, line 158: Something doesn't look right here, but I can't quite tell what's going on. `wb` is being resized while iterating on it?

schancel: Shouldn't `wb.size()` and `wa.size()` be the same on the first pass here?

schancel (changes requested):
Apologies if the comments are silly. My brain feels a bit like mush at the moment.

src/test/monolith_opcodes_tests.cpp
EyeOfPython: Yes, basically we start with `{}` on both `wa` and `wb`, and check the AND/OR/XOR on that, then add one byte to `wb`, check the AND/OR/XOR again, then do the same for all bytes up to 520. Then we add one byte to `wa`, and check all strings `wb` that have equal or longer length again, rinse and repeat until we tested all lengths of `wa` up to 520.

schancel: This code doesn't look like it does that to me. 

1. Start with {}
2. Push back to both `wa` and `wb` (Sizes are the same)
3. Loop for all sizes between `wa.size() and `wb.size() (This doesn't pass the check)
      3.a. Add bytes to `wb` which keeps increasing the size of `wb`
      
It looks like it'd be an infinite loop if it ever actually ran in the first place. Did I miss something?

schancel (approved):

src/script/interpreter.cpp
```diff
@@ -675,35 +675,41 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
-                                    vch1[i] |= vch2[i];
+                                for (size_t i = 0; i < vchShorter.size(); ++i) {
+                                    vchLonger[i] |= vchShorter[i];
                                 }
+                                // leave remaining bytes in vchLonger unchanged
                                 break;
                             case OP_XOR:
-                                for (size_t i = 0; i < vch1.size(); ++i) {
-                                    vch1[i] ^= vch2[i];
+                                for (size_t i = 0; i < vchShorter.size(); ++i) {
```
schancel, line 703: Wait...

schancel (approved):

PR #81, opened 2021-03-11
Implement OP_RAWLEFTBITSHIFT

EyeOfPython:
Replaces `OP_LSHIFT`.

Shifts bits left on positive numbers and right on negative.


src/script/interpreter.cpp
```diff
@@ -1286,6 +1286,92 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
+                        int32_t signed_bitshift =
+                            CScriptNum(stacktop(-1), fRequireMinimal).getint();
+                        popstack(stack);
+                        if (data.empty() || signed_bitshift == 0) {
+                            break;
+                        }
+                        uint32_t bitshift = abs(signed_bitshift);
+                        if (bitshift > data.size() * 8) {
+                            bitshift = data.size() * 8;
+                            //LogPrintf("reset bitshift to %d\n", bitshift);
```
EyeOfPython, line 1305: Code comment

src/script/interpreter.cpp
```diff
@@ -1286,6 +1286,91 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
                         std::reverse(data.begin(), data.end());
                     } break;
 
+                    case OP_RAWBITSHIFT: {
```
schancel, line 1289: Do you have any opinion about inlining these new things? These giant functions drive me a bit mad.

EyeOfPython: I don't know. My C++ isn't great, I mainly imitate the existing style.

BSV has a function at the top of `interpreter.cpp` but it looks very yanky.

schancel: Can leave the style alone for now. But having inline functions would be nice here and in the networking code at some point.

EyeOfPython: Ideally, we'd inline all opcodes and then the interpreter is just
```
case OP_BLAH:
    if (!EvalOpBlah(...)) return false;
    break;
....
```

src/script/interpreter.cpp
```diff
@@ -1286,6 +1285,94 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
+                            if (byteshift < data.size()) {
+                                data[data.size() - byteshift - 1] &=
+                                    0xff << inner_bitshift;
+                            }
+                            for (size_t i = data.size() - byteshift;
+                                    i < data.size(); ++i) {
+                                data[i] = 0;
+                            }
+                        } else { // shift right
+                            uint8_t mask = 0xff << inner_bitshift;
```
schancel, line 1341: I'll review this later.  Some of the same feedback will apply here.

schancel (changes requested):

src/script/interpreter.cpp
```diff
@@ -1286,6 +1285,94 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
+                        uint32_t bitshift = abs(signed_bitshift);
+                        if (bitshift > data.size() * 8) {
+                            bitshift = data.size() * 8;
+                        }
+                        size_t inner_bitshift = bitshift % 8;
+                        size_t byteshift = bitshift / 8;
+                        if (signed_bitshift > 0) {  // shift left
+                            uint8_t mask = 0xff >> inner_bitshift;
+                            for (size_t idx = 0; idx < data.size(); idx++) {
+                                if (idx < byteshift) continue;
```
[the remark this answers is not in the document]
EyeOfPython: I still have to setup `clang-format` 😬

src/script/interpreter.cpp
schancel: Yeah. But you're right. Let's leave it alone for now to avoid backport issues.

src/script/interpreter.cpp
```diff
@@ -1286,6 +1285,95 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
+                        uint32_t bitshift = abs(signed_bitshift);
+                        if (bitshift > data.size() * 8) {
+                            bitshift = data.size() * 8;
+                        }
+                        const size_t inner_bitshift = bitshift % 8;
+                        const size_t byteshift = bitshift / 8;
+                        if (signed_bitshift > 0) { // Shift left
+                            const uint8_t mask = 0xff >> inner_bitshift;
+                            // Copy bits over to the destination
+                            for (size_t idx = byteshift; idx < data.size();
```
schancel, line 1310: Could you drop the hanging conditionals at the end, and instead let this go to `data.size() + byteshift` and then define bits as: `const uint8_t bits = idx < data.size() ? data[idx] : 0;`

src/script/interpreter.cpp
```diff
@@ -1286,6 +1285,95 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
...
+                                 idx++) {
+                                const uint8_t bits = data[idx];
+                                const size_t offset = idx - byteshift;
+                                const uint8_t remove_mask = mask
+                                                            << inner_bitshift;
+                                // Clear existing bits
+                                data[offset] &= ~remove_mask;
+                                // Set new bits
+                                data[offset] |= (bits & mask) << inner_bitshift;
+                                if (offset == 0) {
```
schancel, line 1320: You might want to add another check here for `inner_bitshift == 0` and skip

src/test/op_rawleftbitshift_tests.cpp
```diff
@@ -0,0 +1,847 @@
...
+
+static void CheckTestResultForAllFlags(const stacktype &original_stack,
+                                       const CScript &script,
+                                       const stacktype &expected) {
+    BaseSignatureChecker sigchecker;
+    for (uint32_t flags : flagset) {
+        ScriptError err = ScriptError::OK;
+        stacktype stack{original_stack};
+        bool r = EvalScript(stack, script, flags, sigchecker, &err);
+        BOOST_CHECK(r);
```
schancel, line 39: You should use BOOST_CHECK_MESSAGE here to ensure we can find failing tests in the future.

schancel (approved):

PR #84, opened 2021-03-11
Raise max number of OP_RETURN outputs to 3

### LINKED ISSUE #83: Allow up to 3 OP_RETURN outputs
We don't want to be BSV, but we also want people to innovate. The current limit of just one OP_RETURN output per tx is often seen as very restrictive.

Currently, many protocols use one entire OP_RETURN for payload, following the Lokad ID spec: https://upgradespecs.bitcoincashnode.org/op_return-prefix-guideline/

However, these protocols can interact with each other. Hence, it is useful to be able to use multiple of them.

If we raise the limit of OP_RETURNs to 3, the maximum payload per tx is 660, which is just 140 bytes longer than a single max. push in a scriptSig, so there doesn't seem to be much harm. Theoretically, users could put 1650 bytes per scriptSig.

EyeOfPython:
Closes #83.

### DIFF
--- a/src/test/transaction_tests.cpp
+++ b/src/test/transaction_tests.cpp
@@ -745,7 +745,7 @@ BOOST_AUTO_TEST_CASE(test_IsStandard) {
     t.vout[0].scriptPubKey = CScript() << OP_RETURN;
     BOOST_CHECK(IsStandardTx(CTransaction(t), reason));
 
-    // Only one TxoutType::NULL_DATA permitted in all cases
+    // 2 OP_RETURNs allowed
     t.vout.resize(2);
     t.vout[0].scriptPubKey =
         CScript() << OP_RETURN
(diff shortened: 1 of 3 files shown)


src/test/transaction_tests.cpp
```diff
@@ -745,7 +745,7 @@ BOOST_AUTO_TEST_CASE(test_IsStandard) {
     t.vout[0].scriptPubKey = CScript() << OP_RETURN;
     BOOST_CHECK(IsStandardTx(CTransaction(t), reason));
 
-    // Only one TxoutType::NULL_DATA permitted in all cases
+    // 2 OP_RETURNs allowed
```
schancel, line 748: This might be worth making a loop. But seems fine.

schancel (approved):

src/test/transaction_tests.cpp
EyeOfPython: Without loop, a failure will tell you which test failed

PR #85, opened 2021-03-11
New block header

EyeOfPython:


src/primitives/block.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <hash.h>
 #include <tinyformat.h>
+#include <util/strencodings.h>
 
 BlockHash CBlockHeader::GetHash() const {
     return BlockHash(SerializeHash(*this));
```
schancel, line 13: This will need to be updated at some point for the hashing format you wanted.

src/primitives/block.h
```diff
@@ -22,38 +28,91 @@
...
+
+    SERIALIZE_METHODS(CBlockMetadataField, obj) {
+        READWRITE(obj.nFieldId);
+        READWRITE(obj.vData);
+    }
 };
 
 class CBlock : public CBlockHeader {
 public:
+    std::vector<CBlockMetadataField> vMetadata;
```
schancel, line 115: Shouldn't this be in the header? Or are we doing it as a subtree out of the gate?

schancel (changes requested):
This all makes sense. Is the goal here to just change the header in this commit, and then the hashing algorithm in another commit? I think that would make sense.

EyeOfPython:
I think having the hashing in this commit also makes more sense. I want the tests to pass, but then I'd have to redo all the PoW etc.

We should probably split the metadata part out though. Maybe always enforce it as the hash of '\0' for this PR.

schancel:
Closing, this has been superceeded.

PR #90, opened 2021-03-13
Add 'logos' prefix to coinbase

### LINKED ISSUE #82: Ensure txid are always distinct from Bitcoin Cash and forks
Right now, it's possible to create coinbase txs that have the same hash as txs on Bitcoin Cash, as long as their mined height matches.

This can create quite subtle attack vectors, e.g. creating a tx on Logos thats valid on both chains, sending coins to the victim and then letting them sign a tx that's also valid on BCH, but for higher amounts.

One way to fix this is to pick an additional prefix (prepended to the height) for the coinbase that:
1. Can't be a height
2. and that also occurs in no historic coinbases

For this, we should do the following:
1. Collect all coinbase hex strings
2. Find prefixes that none of the coinbases have and that also isn't a valid height
3. Enforce that prefix on all coinbases

EyeOfPython:
Closes #82.

`test/functional/data/rpc_getblockstats.json` has been generated using the `--gen-test-data` flag.


src/test/miner_tests.cpp
```diff
@@ -196,11 +196,11 @@ void MinerTestingSetup::TestPackageSelection(
     m_node.mempool->addUnchecked(entry.Fee(feeToUse + 2 * SATOSHI).FromTx(tx));
     pblocktemplate = AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey);
     BOOST_CHECK_EQUAL(pblocktemplate->block.vtx.size(), 6);
-    BOOST_CHECK_EQUAL(pblocktemplate->block.vtx[1]->GetId(), lowFeeTxId);
+    BOOST_CHECK_EQUAL(pblocktemplate->block.vtx[1]->GetId(), highFeeTxId);
```
schancel, line 199: We should basically ditch this test. Due to ctor it's nearly irrelevant.

schancel: Well, I'll deal with it once I change the package selection rules.

test/functional/test_framework/blocktools.py
```diff
@@ -60,11 +60,10 @@ def make_conform_to_ctor(block):
 
 def script_BIP34_coinbase_height(height):
     if height <= 16:
-        res = CScriptOp.encode_op_n(height)
-        # Append dummy to increase scriptSig size above 2
-        # (see bad-cb-length consensus rule)
-        return CScript([res, OP_1])
-    return CScript([CScriptNum(height)])
+        num = CScriptOp.encode_op_n(height)
```
schancel, line 63: I hate conditional mutations, but you do you.

schancel (approved):

PR #95, opened 2021-03-18
[backport#19601] Add single sha256 call to CHashWriter

EyeOfPython:
Part 2 of 3.

### DIFF
--- a/src/hash.h
+++ b/src/hash.h
@@ -99,7 +99,7 @@ template <typename T1> inline uint160 Hash160(const T1 &in1) {
 /** A writer stream (for serialization) that computes a 256-bit hash. */
 class CHashWriter {
 private:
-    CHash256 ctx;
+    CSHA256 ctx;
 
     const int nType;
     const int nVersion;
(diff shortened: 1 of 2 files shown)


src/hash.h
```diff
@@ -111,23 +111,38 @@ class CHashWriter {
     int GetVersion() const { return nVersion; }
 
     void write(const char *pch, size_t size) {
-        ctx.Write({(const uint8_t *)pch, size});
+        ctx.Write((const uint8_t *)pch, size);
     }
 
-    // invalidates the object
+    /** Compute the double-SHA256 hash of all data written to this object.
```
schancel, line 117: How is OP_CDS doing the single round of SHA256 right now?

schancel:
```
                        if (vchSig.size()) {
                            valtype vchHash(32);
                            CSHA256()
                                .Write(vchMessage.data(), vchMessage.size())
                                .Finalize(vchHash.data());
                            fSuccess = checker.VerifySignature(
                                vchSig, CPubKey(vchPubKey), uint256(vchHash));
                            metrics.nSigChecks += 1;
```

May want to refactor this as you're doing this. IDK.

EyeOfPython: Great idea.

src/hash.h
```diff
@@ -112,23 +112,38 @@ class CHashWriter {
     int GetVersion() const { return nVersion; }
 
     void write(const char *pch, size_t size) {
-        ctx.Write({(const uint8_t *)pch, size});
+        ctx.Write((const uint8_t *)pch, size);
```
schancel, line 115: Why does this need to change? Seems like an optimization of some sort?

src/hash.h
```diff
@@ -112,23 +112,38 @@ class CHashWriter {
...
 
-    // invalidates the object
+    /** Compute the double-SHA256 hash of all data written to this object.
+     *
+     * Invalidates this object.
+     */
     uint256 GetHash() {
         uint256 result;
-        ctx.Finalize(result);
+        ctx.Finalize(result.begin());
```
schancel, line 124: Why does this need to change? Is it just to reset it?

schancel (approved):

src/hash.h
EyeOfPython: No, they just have different signatures:

```
CHash256 &Write(Span<const uint8_t> input);
vs
CSHA256 &Write(const uint8_t *data, size_t len);
```

🤷‍♂️

src/hash.h
EyeOfPython: Again, different signatures:

```
void CHash256::Finalize(Span<uint8_t> output);
vs.
void CSHA256::Finalize(uint8_t hash[32]);
```

Should probably be refactored at some point, but not yet.

Issue #88, opened 2021-03-12
Implement BTC's Taproot (BIP341)


EyeOfPython:
Backporting a modified version of Taproot ([BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)) would give us a bunch of advantages:

* MAST, which allows representing scripts with a lot of IFs in a much more efficient and private way.
   * This also on *some* level increases the opcode limit by a factor of 2^128, which is the number of possible leaves. With proper CashScript support, this would make writing smart contracts a lot easier, as developers don't have to check whether their script broke the limit as much.
* Spending P2SH-esque outputs without revealing the script, adding privacy for complex scripts, especially useful for those involved in business banned in some countries, e.g. gambling.
* Advanced txs insight through a new sighash algorithm, allowing introspecting the amounts and `scriptPubKey`s of other inputs.
  * Also allows much more efficient introspection as parts of the serialization are simply left out if they're not part of the sig hash type.
* Wallet developers could be encouraged to always use Taproot but with an empty (i.e. invalid) root and then just always spend it directly, making payments a little bit more efficient and all unrevealed smart contract payments much more private.
* If we don't change the sighash algorithm (too much), it would make it easier for multicoin hardware wallets that support taproot to also support Logos.
* If we add `OP_ECADD` and `OP_ECMUL`, smart contract developers can still use recurrent smart contracts in the same way they currently do.

I would suggest modifying BIP341 in the following way:
* Add a dedicated opcode, OP_TAPROOT, which always fails script execution immediately.
* In VerifyScript, if a `scriptPubKey` has the form `<32 bytes> OP_TAPROOT`, instead of executing it, run the taproot verification in its place.
* Instead of using the witness stack, use the normal stack from scriptSig.
* Add FORKID 0x40 to the `hash_type`.

EyeOfPython:
This has been implemented.

PR #101, opened 2021-03-21
Add ScriptExecutionData and thread it through EvalScript.

EyeOfPython:
This is required to collect the data for the BIP341 sighash, which, among other things, contains the last executed codeseparator position.

Required for #88.


schancel:
Will this end up getting cached? Why is it included in the sighash?

EyeOfPython:
The SIGHASH is not getting cached. All that is cached is in PrecomputedTransactionData.

In BIP143, the codeseparator makes the scriptCode in the sighash cut out the script bytes before (and including) it.

In BIP341+342, it instead just keeps track of the codeseparator position as an index, significantly reducing the max. number of hashed bytes.

It’s just a different way to express the sighash.

schancel (approved):

PR #102, opened 2021-03-21
Fix bench-bitcoin bug in BlockToJsonVerbose benchmark.

EyeOfPython:
62c2bc5 added the requirement to check LowS for verbose serialization of sigs, but we didn't fix the benchmarks.

### DIFF
--- a/src/bench/rpc_blockchain.cpp
+++ b/src/bench/rpc_blockchain.cpp
@@ -5,6 +5,7 @@
 #include <bench/bench.h>
 #include <bench/data.h>
 
+#include <pubkey.h>
 #include <rpc/blockchain.h>
 #include <streams.h>
 #include <validation.h>
@@ -12,6 +13,7 @@
 #include <univalue.h>
 
 static void BlockToJsonVerbose(benchmark::Bench &bench) {
+    const ECCVerifyHandle verify_handle;
     CDataStream stream(benchmark::data::block413567, SER_NETWORK,
                        PROTOCOL_VERSION);
     char a = '\0';


src/bench/rpc_blockchain.cpp
```diff
@@ -5,13 +5,15 @@
...
 
+#include <pubkey.h>
 #include <rpc/blockchain.h>
 #include <streams.h>
 #include <validation.h>
 
 #include <univalue.h>
 
 static void BlockToJsonVerbose(benchmark::Bench &bench) {
+    const ECCVerifyHandle verify_handle;
```
schancel, line 16: What exactly was the bug and how does simply defining this fix it?

EyeOfPython: Yes.

So since we always enforce Low S, the signature printing (for the human readable script asm) needs to check if the signature has low S.

For that, we need access to a secp256k1 verify context.

ECCVerifyHandle sets this up for us in the constructor and tears it down again in the destructor.

schancel: Can you add that to the commit message? There's "spooky" action at a distance here. This doesn't seem to be passed in anywhere.

schancel (approved):

src/bench/rpc_blockchain.cpp
EyeOfPython: Yeah, who doesn’t love globals? I ofc would pass the context down the functions, but in bitcoind it’s just a global defined somewhere 🤷🏼‍♂️

I can add it in the commit message.

PR #103, opened 2021-03-21
Add OP_MULPOW2.

EyeOfPython:
This computes x n -> x * 2^n, which is useful for many applications.


src/script/interpreter.cpp
```diff
@@ -93,7 +93,6 @@ static bool IsOpcodeDisabled(opcodetype opcode, uint32_t flags) {
         case OP_2MUL:
         case OP_2DIV:
         case OP_MUL:
-        case OP_RSHIFT:
```
schancel, line 96: Why's this going away?

schancel: Nevermind

EyeOfPython: 😉

src/script/script.h
```diff
@@ -362,6 +362,34 @@ class CScriptNum {
         return *this;
     }
 
+    inline CScriptNum mulpow2(const CScriptNum rhs) {
+        const int64_t sign = m_value > 0 ? 1 : -1;
+        // We use unsigned integers for shifts, they have well-defined behavior
+        const uint64_t absval = m_value > 0 ? m_value : -m_value;
+        if (absval == 0) {
+            return CScriptNum(0);
```
schancel, line 370: You can move this up first. Saves two branches in this case.

src/script/script.h
```diff
@@ -362,6 +362,34 @@ class CScriptNum {
         return *this;
     }
 
+    inline CScriptNum mulpow2(const CScriptNum rhs) {
+        const int64_t sign = m_value > 0 ? 1 : -1;
+        // We use unsigned integers for shifts, they have well-defined behavior
+        const uint64_t absval = m_value > 0 ? m_value : -m_value;
```
schancel, line 368: Should note that this is defined as `x` for reference below.

schancel (changes requested):

src/test/op_mulpow2_tests.cpp
```diff
@@ -0,0 +1,137 @@
...
+        bool r = EvalScript(stack, script, flags, sigchecker, &err);
+        BOOST_CHECK_MESSAGE(
+            !r, "Expected " << ScriptErrorString(expected_error) << " but got "
+                            << ScriptErrorString(err) << " for '"
+                            << ScriptToAsmStr(script) << "'");
+        BOOST_CHECK_EQUAL(err, expected_error);
+    }
+}
+
+static bool Pow2(int64_t base, int64_t exponent, int64_t &result) {
```
schancel, line 78: I would consider using the existing pow math functions. But this also works.

schancel (changes requested):

schancel (approved):

PR #107, opened 2021-03-26
Fix broken tests: SignSignature is called for tx with invalid inputs

EyeOfPython:
In a few places, SignSignature is called with transactions that have inputs that don't have a corresponding output.

- In DoS_mapOrphans, adds 2777 inputs to `tx`, but doesn't add the corresponding outputs to `txPrev`.
- In AreInputsStandard, someone typed 5 instead of 4.
- In script_combineSigs, `txFrom` is updated, but `txTo` doesn't have its corresponding prevout updated.


schancel (approved):
How come these don’t fail right now?

EyeOfPython:
Because it wasn't tested for

I was implementing the BIP341 sighash stuff and then these tests started to fail

PR #110, opened 2021-03-27
Make SignatureHash fallible

EyeOfPython:
It now returns bool instead of uint256, and gets the output as parameter.

This is required for BIP341, because sighash calculation can fail there.


schancel (approved):

schancel (changes requested):
Needs rebase now

schancel (approved):

PR #112, opened 2021-03-27
Refactor: thread ScriptExecutionData to SignatureHash.

EyeOfPython:
ScriptExecutionData contains `codeseparator_pos` and `executed_script_hash`, which is required for the BIP341 sighash.

### DIFF
--- a/src/script/script_exec_data.h
+++ b/src/script/script_exec_data.h
@@ -5,6 +5,8 @@
 #ifndef BITCOIN_SCRIPT_SCRIPT_EXEC_DATA_H
 #define BITCOIN_SCRIPT_SCRIPT_EXEC_DATA_H
 
+#include <hash.h>
+#include <script/script.h>
 #include <uint256.h>
 
 /**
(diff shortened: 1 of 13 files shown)


src/script/script_exec_data.h
```diff
@@ -5,19 +5,48 @@
...
+     * OP_CODESEPARATOR.
+     *
+     * - In plain scripts, this is the SHA256 of the scriptPubKey.
+     * - In P2SH, this is the SHA256 of the redeemScript.
+     * - In Tapscripts, this is the "tapleaf hash", calculated as defined in
+     *   BIP341: taggedhash_TapLeaf(version || compact_size(len(script)) ||
+     *                              script).
+     * - In Taproot, this entire struct would not be constructed.
+     */
+    std::optional<uint256> m_executed_script_hash;
```
schancel, line 37: Should this be optional?

src/script/script_exec_data.h
```diff
@@ -5,19 +5,48 @@
...
+
+    /**
+     * ScriptExecutionData with the SHA-256 of executed_script and as
+     * codeseparator position.
+     */
+    static ScriptExecutionData
+    SHA256Script(const CScript &executed_script,
```
schancel, line 44: Does this relate to `m_executed_script_hash` above?

src/script/script_exec_data.h
```diff
@@ -5,19 +5,48 @@
...
+                 uint32_t codeseparator_pos = DEFAULT_CODESEP_POS) {
+        CHashWriter ss(SER_GETHASH, 0);
+        ss << executed_script;
+        return {DEFAULT_CODESEP_POS, ss.GetSHA256()};
```
schancel, line 48: Should this not use `codeseparator_pos`?

src/script/script_exec_data.h
schancel: I think we should define a non-default constructor here, and use this in the constructor.  Require a `CScript`, but it would just be empty for the initial case of executing the scriptSig.

schancel: This will prevent the inadvertant use here without defining the m_executed_script_hash.

schancel (approved):

PR #116, opened 2021-03-29
Add BIP341 sighash support to bitcoin-tx. 

EyeOfPython:
Now `sign=x|BIP341|FORKID` and `sign=x|BIP341|FORKID|ANYONECANPAY` can be used.

Note: The BIP341 signatures are still invalid due to `SigHashType::isDefined` returning false, therefore the tests have signatures ending in `6x` or `ex` instead of `[x|BIP341|FORKID]` etc.
This will need to be changed once SigHashType::isDefined allows BIP341 sigs.


src/script/sign.cpp
```diff
@@ -19,7 +19,7 @@ MutableTransactionSignatureCreator::MutableTransactionSignatureCreator(
     const Amount &amountIn, SigHashType sigHashTypeIn,
     const PrecomputedTransactionData &txdata)
     : txTo(txToIn), nIn(nInIn), amount(amountIn), sigHashType(sigHashTypeIn),
-      checker(txTo, nIn, amountIn, txdata) {}
+      checker(txTo, nIn, amountIn, txdata), m_txdata(txdata) {}
```
schancel, line 22: Why did this need to change? Is this a bug in a previous commit?

schancel (approved):

PR #118, opened 2021-03-30
Fix bitcoin-tx tests.

EyeOfPython:
They currently don't sign the output (sign first, then add output) and use the legacy sighash.

Now they add output first and sign then, and use the FORKID sighash.
Also add a SINGLE test, which would produce an empty signature if signed in the wrong order as before.


schancel (approved):
We should probably offer PiRK this commit

PR #119, opened 2021-03-30
Rename OP_VER to OP_SCRIPTTYPE, add INVALID_OP_SCRIPTTYPE.

EyeOfPython:
This marker opcode allows us to introduce new models of execution, such as Taproot.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -83,7 +83,6 @@ static void CleanupScriptCode(CScript &scriptCode,
 static bool IsOpcodeDisabled(opcodetype opcode, uint32_t flags) {
     switch (opcode) {
         case OP_RESERVED:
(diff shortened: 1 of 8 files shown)


src/script/interpreter.cpp
```diff
@@ -263,6 +262,10 @@ bool EvalScript(std::vector<valtype> &stack, const CScript &script,
                 return set_error(serror, ScriptError::DISABLED_OPCODE);
             }
 
+            if (opcode == OP_SCRIPTTYPE) {
```
schancel, line 265: You're setting it up to delete this in a future commit where taproot is enabled?

schancel: Nevermind. This is going to be a pattern.

schancel (approved):

PR #120, opened 2021-03-30
Add CPubKey::VerifyPubkeyTweakAdd.

EyeOfPython:
This verifies that the pubkey is the base point tweaked by a tweak; in other words base + tweak*G.

This is required for Taproot, for verifying that the hash of the tapleaf combined with the base point results in the commitment.


src/pubkey.cpp
```diff
@@ -225,6 +225,29 @@ bool CPubKey::VerifySchnorr(const uint256 &hash,
     return VerifySchnorr(hash, sig);
 }
 
+bool CPubKey::VerifyPubkeyTweakAdd(const CPubKey &base,
```
schancel, line 228: Why not return the tweaked value?

```
bool CPubKey::VerifyPubkeyTweakAdd(const CPubKey &base, const uint256 &tweak, CPubKey *out = nullptr) const {

if(out) {
    *out = Whatever is needed
}
```

EyeOfPython: Rename to `AddScalar`

PR #126, opened 2021-04-03
Add IsPayToTaproot

EyeOfPython:
This returns whether a script is a valid Taproot output script.

Valid scripts can be either of these:
- `OP_SCRIPTTYPE OP_1 <33-byte commitment>
- `OP_SCRIPTTYPE OP_1 <33-byte commitment> <32-byte state>

### DIFF
--- a/src/script/taproot.cpp
+++ b/src/script/taproot.cpp
@@ -4,6 +4,7 @@
 
 #include <hash.h>
 #include <pubkey.h>
+#include <script/script.h>
 #include <script/taproot.h>
 
 static const CHashWriter HASHER_TAPLEAF = TaggedHash("TapLeaf");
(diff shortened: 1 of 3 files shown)


src/script/taproot.cpp
```diff
@@ -62,3 +63,24 @@ bool VerifyTaprootCommitment(uint256 &tapleaf_hash,
...
+    // First push (commitment) must be 33 bytes long
+    if (script[2] != CPubKey::COMPRESSED_SIZE) {
+        return false;
+    }
+    // If there's only a commitment but no state, we're valid
+    if (script.size() == TAPROOT_SIZE_WITHOUT_STATE) {
+        return true;
+    }
+    // Otherwise, we need a state with 32 bytes.
+    return script.size() == TAPROOT_SIZE_WITH_STATE &&
```
schancel, line 84: Are we going to be allowing state in the first iteration?

EyeOfPython: I’d say so, otherwise the smart contract capabilities are severely limited. We’d need EC_MUL and EC_ADD to do recursive smart contracts

schancel (approved):

PR #128, opened 2021-04-03
Add SCRIPT_REQUIRE_TAPROOT_SIGNATURES

EyeOfPython:
Ensures that signatures are Schnorr signatures (64 byte long w/o sig hash flag) and use the BIP341 sighash.

This flag is required for verifying signatures for the Taproot key spend path (spending without revealing the script).

Is is never set during script execution, only during signature verification.

### DIFF
--- a/src/script/script_flags.h
+++ b/src/script/script_flags.h
@@ -11,6 +11,11 @@
 enum {
     SCRIPT_VERIFY_NONE = 0,
 
+    // Require BIP341 sighash and Schnorr signatures for key spend path
+    // (spending without revealing the script) in Taproot.
+    // Only used for checking signatures, not during script execution.
+    SCRIPT_REQUIRE_TAPROOT_SIGNATURES = (1U << 0),
+
     // Discourage use of NOPs reserved for upgrades (NOP1-10)
     //
     // Provided so that nodes can avoid accepting or mining transactions
--- a/src/test/checkdatasig_tests.cpp
+++ b/src/test/checkdatasig_tests.cpp
@@ -140,6 +140,8 @@ BOOST_AUTO_TEST_CASE(checkdatasig_test) {
     MMIXLinearCongruentialGenerator lcg;
     for (int i = 0; i < 4096; i++) {
         uint32_t flags = lcg.next();
+        // Flag cannot occur in script execution, only in Taproot key spend
+        flags &= ~SCRIPT_REQUIRE_TAPROOT_SIGNATURES;
 
         // Since strict encoding is enforced, hybrid keys are invalid.
         CheckError(flags, {{}, message, pubkeyH}, script,
--- a/src/test/sigencoding_tests.cpp
+++ b/src/test/sigencoding_tests.cpp
@@ -21,10 +21,16 @@ static valtype SignatureWithHashType(valtype vchSig, SigHashType sigHash) {
 static void CheckSignatureEncodingWithSigHashType(const valtype &vchSig,
                                                   uint32_t flags) {
     ScriptError err = ScriptError::OK;
-    BOOST_CHECK(CheckDataSignatureEncoding(vchSig, flags, &err));
 
     const bool hasForkId = (flags & SCRIPT_ENABLE_SIGHASH_FORKID) != 0;
+    // isTaproot requires Schnorr sigs and BIP341 sighash
+    const bool isTaproot = (flags & SCRIPT_REQUIRE_TAPROOT_SIGNATURES) != 0;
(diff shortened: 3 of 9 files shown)


schancel:
Can you add the rationale in the message? I don't understand this. I thought we were allowing Taproot sigs everywhere?

EyeOfPython:
Yes we do. We just don’t allow FORKID/BIP143 for the key spend path.
The BIP341 sighash was designed specifically for taproot. We extended it to P2SH and P2PKH, to allow other interesting smart contract schemes.

src/script/script_flags.h
```diff
@@ -48,6 +48,10 @@ enum {
     // Note: The Segwit Recovery feature is a (currently moot) exception to
     // VERIFY_INPUT_SIGCHECKS
     SCRIPT_VERIFY_INPUT_SIGCHECKS = (1U << 22),
+
+    // Require BIP341 sighash for key spend path (spending without revealing the
+    // script) in Taproot
+    SCRIPT_REQUIRE_BIP341_SIGHASH = (1U << 23),
```
schancel, line 54: May want to move this down a bit since ABC may use this bit, though it's easy enough to rearrange theirs in a backport.

schancel (approved):

src/test/checkdatasig_tests.cpp
```diff
@@ -140,6 +140,8 @@ BOOST_AUTO_TEST_CASE(checkdatasig_test) {
     MMIXLinearCongruentialGenerator lcg;
     for (int i = 0; i < 4096; i++) {
         uint32_t flags = lcg.next();
+        // Flag cannot occur in script execution, only in Taproot key spend
+        flags &= ~SCRIPT_REQUIRE_TAPROOT_SIGNATURES;
```
schancel, line 144: Seems like there should be a test that this fails?

src/test/sigencoding_tests.cpp
```diff
@@ -21,11 +21,17 @@ static valtype SignatureWithHashType(valtype vchSig, SigHashType sigHash) {
...
                                                   uint32_t flags) {
     ScriptError err = ScriptError::OK;
-    BOOST_CHECK(CheckDataSignatureEncoding(vchSig, flags, &err));
 
     const bool hasForkId = (flags & SCRIPT_ENABLE_SIGHASH_FORKID) != 0;
+    // isTaproot requires Schnorr sigs and BIP341 sighash
+    const bool isTaproot = (flags & SCRIPT_REQUIRE_TAPROOT_SIGNATURES) != 0;
     const bool is64 = (vchSig.size() == 64);
 
+    // isTaproot is never true for OP_CHECKDATASIG, but we test it anyway to
```
schancel, line 30: Can you elaborate on this?

src/test/sigencoding_tests.cpp
```diff
@@ -47,13 +53,18 @@ static void CheckSignatureEncodingWithSigHashType(const valtype &vchSig,
     }
 
     for (const SigHashType &sigHash : sigHashes) {
-        // Check the signature with the proper forkid flag.
+        // isTaproot requires Schnorr sigs and BIP341 sighash
+        const bool expectedValid =
+            isTaproot ? is64 && sigHash.hasBIP341() : true;
```
schancel, line 58: This is basically requiring that if it's a BIP341 sighash then it must be a schnorr signature? Is this true in OP_CHECKSIG?

src/test/checkdatasig_tests.cpp
EyeOfPython: It's tested in e.g. sigencoding_tests.cpp, so here it seemed superfluous, especially since it's against the "spirit" of the test.

src/test/sigencoding_tests.cpp
EyeOfPython: Hmm, so we use the `SCRIPT_REQUIRE_TAPROOT_SIGNATURES` script flag to enforce BIP341 and Schnorr signatures for the key spend path, which is only ever set for that sig check (using `EvalChecksig`).

Maybe changing it to `isTRKeySpend` helps? It shouldn't be too long otherwise the predicates become unreadable.

I could also change the flag to `SCRIPT_TAPROOT_KEY_SPEND_PATH`, do you think that would be more clear?

src/test/sigencoding_tests.cpp
EyeOfPython: No, `isTaproot` is only true for directly spending a taproot using a single signature. Possibly it should be renamed to `isTRKeySpend`

EyeOfPython:
Superseded by #131

PR #129, opened 2021-04-03
Add VerifyScriptType and VerifyTaprootSpend stub

EyeOfPython:
Introduces VerifyScriptType, which runs in stead of the scriptPubKey (and potential following P2SH).

For script type 1, VerifyTaprootSpend gets run, which currently just checks IsPayToTaproot and then returns UNKNOWN error.


src/script/interpreter.cpp
```diff
@@ -2063,55 +2081,69 @@ bool VerifyScript(const CScript &scriptSig, const CScript &scriptPubKey,
             return false;
         }
     }
-    stackCopy = stack;
```
EyeOfPython, line 2066: The diff is super messy, but it esentially just indents the `else` branch.

src/script/interpreter.cpp
```diff
@@ -2063,55 +2093,69 @@ bool VerifyScript(const CScript &scriptSig, const CScript &scriptPubKey,
...
-        assert(!stack.empty());
-
-        const valtype &pubKeySerialized = stack.back();
-        CScript pubKey2(pubKeySerialized.begin(), pubKeySerialized.end());
-        popstack(stack);
-
+        if (!CastToBool(stack.back())) {
+            return set_error(serror, ScriptError::EVAL_FALSE);
+        }
+    } else {
```
schancel, line 2108: This else isn't needed and just causes indentation.

src/script/interpreter.cpp
```diff
@@ -2063,55 +2093,69 @@ bool VerifyScript(const CScript &scriptSig, const CScript &scriptPubKey,
...
-    if (CastToBool(stack.back()) == false) {
-        return set_error(serror, ScriptError::EVAL_FALSE);
-    }
-
-    // Additional validation for spend-to-script-hash transactions:
-    if (scriptPubKey.IsPayToScriptHash()) {
-        // scriptSig must be literals-only or validation fails
-        if (!scriptSig.IsPushOnly()) {
-            return set_error(serror, ScriptError::SIG_PUSHONLY);
+        if (stack.empty()) {
```
schancel, line 2102: Not checking cleanstack here.

src/script/interpreter.cpp
```diff
@@ -2063,55 +2093,69 @@ bool VerifyScript(const CScript &scriptSig, const CScript &scriptPubKey,
...
-        // EvalScript above would return false.
-        assert(!stack.empty());
-
-        const valtype &pubKeySerialized = stack.back();
-        CScript pubKey2(pubKeySerialized.begin(), pubKeySerialized.end());
-        popstack(stack);
-
+        if (!CastToBool(stack.back())) {
+            return set_error(serror, ScriptError::EVAL_FALSE);
+        }
```
schancel, line 2107: just return true?

src/script/interpreter.cpp
EyeOfPython: Yes, it's checked here: https://github.com/cashweb/cashd/pull/129/files#diff-a0337ffd7259e8c7c9a7786d6dbd420c80abfa1afdb34ebae3261109d9ae3c19R2165

src/script/interpreter.cpp
EyeOfPython: Actually we need it cuz after the else we check for cleanstack and sigchecks and setting the metrics.

schancel (approved):
My only concern is all the double and triple checks of the script format due to the way things are structured. I would almost prefer `IsTapRootSpend` to wrap a function like `GetScriptVersion.` Or even some kind of polymorphic class to execute scripts.  Repeating the necessary conditions for spending in multiple places is error prone, though if we write good tests it should be okay.

PR #130, opened 2021-04-06
Add Taproot key spend path

EyeOfPython:
This allows Taproot outputs to be spent with a signature for the commitment using BIP341 signatures.


schancel:
Looks pretty straightforward. Let me go through it with a fine tooth comb a bit later.

src/script/interpreter.cpp
```diff
@@ -2051,7 +2051,38 @@ static bool VerifyTaprootSpend(std::vector<valtype> stack,
...
+                          script_pubkey.end(),
+                          flags | SCRIPT_TAPROOT_KEY_SPEND_PATH, checker,
+                          metrics, std::nullopt, serror, f_success)) {
+            if (serror && *serror == ScriptError::SIG_NULLFAIL) {
+                *serror = ScriptError::TAPROOT_VERIFY_SIGNATURE_FAILED;
+            }
+            // serror is set
+            return false;
+        }
+        if (!f_success) {
```
schancel, line 2079: Might just say that this is the case when vch_sig is 0.

src/script/interpreter.cpp
```diff
@@ -2051,7 +2051,38 @@ static bool VerifyTaprootSpend(std::vector<valtype> stack,
...
+    }
+    if (stack.size() >= 2 && !stack.back().empty() &&
+        stack.back()[0] == TAPROOT_ANNEX_TAG) {
+        return set_error(serror, ScriptError::TAPROOT_ANNEX_NOT_SUPPORTED);
+    }
+    if (stack.size() == 1) {
+        // Spend using single signature instead of executing script
+        valtype &vch_sig = stacktop(-1);
+        bool f_success = false;
+        if (!EvalChecksig(vch_sig, vch_pubkey, script_pubkey.begin(),
```
schancel, line 2069: I would almost check for null here just to be sure.

schancel:
```C++
if (!vchSig.size()) {
  return set_error(serror,ScriptError::TAPROOT_VERIFY_SIGNATURE_FAILED);
}
...
assert(f_success);
```

schancel: My only concern with this, is that we don't test something and then end up with the assert in a place where a remote crash can happen.

schancel (approved):

src/script/interpreter.cpp
```diff
@@ -2051,7 +2051,34 @@ static bool VerifyTaprootSpend(std::vector<valtype> stack,
...
+        }
+        if (vch_sig.empty() || !checker.CheckSig(vch_sig, vch_pubkey,
+                                                 std::nullopt, {}, sig_flags)) {
+            return set_error(serror,
+                             ScriptError::TAPROOT_VERIFY_SIGNATURE_FAILED);
+        }
+        return set_success(serror);
+    }
+    // Temporarily return UNKNOWN for script spend path
     return set_error(serror, ScriptError::UNKNOWN);
```
schancel, line 2082: This ends up having a fall through on non-annex based stack.size() of >=2

PR #131, opened 2021-04-10
Add SCRIPT_TAPROOT_KEY_SPEND_PATH script flag

EyeOfPython:
Ensures that signatures are Schnorr signatures (64 byte long + sig hash flag) and use the BIP341 sighash.

This flag is required for verifying signatures for the Taproot key spend path (spending without revealing the script).

It is never set during script execution, only during signature verification.

### DIFF
--- a/src/script/script_error.h
+++ b/src/script/script_error.h
@@ -89,6 +89,11 @@ enum class ScriptError {
        marker */
     INVALID_OP_SCRIPTTYPE,
 
+    /* Key spend path for Taproot must use BIP341 sighash */
+    TAPROOT_KEY_SPEND_MUST_USE_BIP341_SIGHASH,
+    /* Key spend path for Taproot must Schnorr signatures */
+    TAPROOT_KEY_SPEND_MUST_USE_SCHNORR_SIG,
+
     ERROR_COUNT,
 };
 
--- a/src/test/checkdatasig_tests.cpp
+++ b/src/test/checkdatasig_tests.cpp
@@ -140,6 +140,8 @@ BOOST_AUTO_TEST_CASE(checkdatasig_test) {
     MMIXLinearCongruentialGenerator lcg;
     for (int i = 0; i < 4096; i++) {
         uint32_t flags = lcg.next();
+        // Flag cannot occur in script execution, only in Taproot key spend
+        flags &= ~SCRIPT_TAPROOT_KEY_SPEND_PATH;
 
         // Since strict encoding is enforced, hybrid keys are invalid.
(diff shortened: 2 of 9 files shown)


src/test/checkdatasig_tests.cpp
```diff
@@ -140,6 +140,8 @@ BOOST_AUTO_TEST_CASE(checkdatasig_test) {
     MMIXLinearCongruentialGenerator lcg;
     for (int i = 0; i < 4096; i++) {
         uint32_t flags = lcg.next();
+        // Flag cannot occur in script execution, only in Taproot key spend
+        flags &= ~SCRIPT_TAPROOT_KEY_SPEND_PATH;
```
schancel, line 144: Would it make sense to test that this fails at the end of this loop? or does it get tested somewhere else?

src/script/script_error.h
```diff
@@ -89,6 +89,11 @@ enum class ScriptError {
        marker */
     INVALID_OP_SCRIPTTYPE,
 
+    /* Key spend path for Taproot must use BIP341 sighash */
```
schancel, line 92: I would just expand this comment, and use doc comments here. Note that it's prohibited from EvalScript, but is used to pass through given the existing flag infrastructure for CheckSig* stuff.

src/test/sigencoding_tests.cpp
```diff
@@ -21,11 +21,17 @@ static valtype SignatureWithHashType(valtype vchSig, SigHashType sigHash) {
 static void CheckSignatureEncodingWithSigHashType(const valtype &vchSig,
                                                   uint32_t flags) {
     ScriptError err = ScriptError::OK;
-    BOOST_CHECK(CheckDataSignatureEncoding(vchSig, flags, &err));
 
     const bool hasForkId = (flags & SCRIPT_ENABLE_SIGHASH_FORKID) != 0;
+    // isTRKeySpend requires Schnorr sigs and BIP341 sighash
```
schancel, line 26: Due to the piping for CDS, the `SCRIPT_ENABLE_SIGHASH_FORKID` enforces schnorr always as a side effect. This impacts CDS functionality (in tests) since the opcode script cannot be executed with this flag.

schancel (approved):

src/script/script_error.h
EyeOfPython: I think you mean `SCRIPT_TAPROOT_KEY_SPEND_PATH`, not these errors?

schancel: image

PR #132, opened 2021-04-10
Add script spend path for Taproot

EyeOfPython:
This allows spending an output by revealing the interal pubkey, a leaf script and a merkle path proving the commitment.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -2042,6 +2042,45 @@ bool GenericTransactionSignatureChecker<T>::CheckSequence(
 template class GenericTransactionSignatureChecker<CTransaction>;
 template class GenericTransactionSignatureChecker<CMutableTransaction>;
 
+bool VerifyScriptPostConditions(const std::vector<valtype> stack,
+                                const CScript &scriptSig, uint32_t flags,
+                                const ScriptExecutionMetrics &metrics,
+                                ScriptError *serror) {
+    // The CLEANSTACK check is only performed after potential P2SH evaluation,
+    // as the non-P2SH evaluation of a P2SH script will obviously not result in
+    // a clean stack (the P2SH inputs remain).
+    if ((flags & SCRIPT_VERIFY_CLEANSTACK) != 0) {
+        if (stack.size() != 1) {
+            return set_error(serror, ScriptError::CLEANSTACK);
+        }
+    }
+
+    if (flags & SCRIPT_VERIFY_INPUT_SIGCHECKS) {
+        // This limit is intended for standard use, and is based on an
+        // examination of typical and historical standard uses.
+        // - allowing P2SH ECDSA multisig with compressed keys, which at an
+        // extreme (1-of-15) may have 15 SigChecks in ~590 bytes of scriptSig.
+        // - allowing Bare ECDSA multisig, which at an extreme (1-of-3) may have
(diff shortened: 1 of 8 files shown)


src/script/interpreter.cpp
```diff
@@ -2041,6 +2042,148 @@ bool GenericTransactionSignatureChecker<T>::CheckSequence(
...
+        // Spend using single signature instead of executing script
+        valtype &vch_sig = stacktop(-1);
+        bool f_success = false;
+        if (!EvalChecksig(vch_sig, vch_pubkey, script_pubkey.begin(),
+                          script_pubkey.end(),
+                          flags | SCRIPT_TAPROOT_KEY_SPEND_PATH, checker,
+                          metrics, std::nullopt, serror, f_success)) {
+            if (serror && *serror == ScriptError::SIG_NULLFAIL) {
+                *serror = ScriptError::TAPROOT_VERIFY_SIGNATURE_FAILED;
+            }
```
schancel, line 2114: Is cleanstack something we really want to make optional in taproot?

src/script/interpreter.cpp
```diff
@@ -2041,6 +2042,148 @@ bool GenericTransactionSignatureChecker<T>::CheckSequence(
...
 template class GenericTransactionSignatureChecker<CMutableTransaction>;
 
+bool VerifyScriptPostConditions(const std::vector<valtype> stack,
+                                const CScript &scriptSig, uint32_t flags,
+                                const ScriptExecutionMetrics &metrics,
+                                ScriptError *serror) {
+    // The CLEANSTACK check is only performed after potential P2SH evaluation,
+    // as the non-P2SH evaluation of a P2SH script will obviously not result in
+    // a clean stack (the P2SH inputs remain).
+    if ((flags & SCRIPT_VERIFY_CLEANSTACK) != 0) {
```
schancel, line 2052: We don't really need this for taproot execution, right? We could just check the stack size along with bool.

src/script/interpreter.cpp
```diff
@@ -2041,6 +2042,148 @@ bool GenericTransactionSignatureChecker<T>::CheckSequence(
...
+            valtype(script_pubkey.begin() + TAPROOT_SIZE_WITHOUT_STATE + 1,
+                    script_pubkey.begin() + TAPROOT_SIZE_WITH_STATE));
+    }
+    ScriptExecutionData execdata{tapleaf_hash};
+    if (!EvalScript(stack, exec_script, flags, checker, metrics, execdata,
+                    serror)) {
+        // serror is set
+        return false;
+    }
+    if (stack.empty() || CastToBool(stack.back()) == false) {
```
schancel, line 2158: You're already checking if it's not empty here, why not check it above and set cleanstack?

src/script/interpreter.cpp
```diff
@@ -2041,6 +2042,148 @@ bool GenericTransactionSignatureChecker<T>::CheckSequence(
...
+        return set_error(serror, ScriptError::EVAL_FALSE);
+    }
+    if (!VerifyScriptPostConditions(stack, script_sig, flags, metrics,
```
schancel, line 2161: I get the sigcheck asserts, however it would be better to maintain a counter in the EvalScript metrics and fail prematurely rather than continuing to execute them. I suppose this is okay as we won't be able to get spammed with failing transactions though.

schancel (approved):

PR #133, opened 2021-04-10
Add Taproot key spend path

EyeOfPython:
This allows Taproot outputs to be spent with a signature for the commitment using BIP341 signatures.

They are still non-standard, but will become standard in a future commit


src/script/interpreter.cpp
```diff
@@ -2051,7 +2051,36 @@ static bool VerifyTaprootSpend(std::vector<valtype> stack,
...
+        // Spend using single signature instead of executing script
+        const valtype &vch_sig = stacktop(-1);
+        const uint32_t sigflags = flags | SCRIPT_TAPROOT_KEY_SPEND_PATH;
+        if (!CheckTransactionSignatureEncoding(vch_sig, sigflags, serror) ||
+            !CheckPubKeyEncoding(vch_pubkey, sigflags, serror)) {
+            // serror is set
+            return false;
+        }
+
+        if (vch_sig.size() > 0 &&
```
schancel, line 2075: This seems fine, I'd still for verbosity and clarity move the `if(vch_sig.size() == 0) { return set_error(...}` here as well.

schancel:
```C++
if(vch_sig.size() == 0) { 
    return set_error(...}
}
const boolean fSuccess = checker.CheckSig(vch_sig, vch_pubkey, std::nullopt, {}, sigflags);
if(!fSuccess) {
   return set_error(serror,
                             ScriptError::TAPROOT_VERIFY_SIGNATURE_FAILED);
}
return set_success(serror);
```

schancel (approved):

schancel (approved):

PR #144, opened 2021-04-12
Make non-pushonly scriptsig in coinbase inputs standard

EyeOfPython:
Currently, non-pushonly scriptSigs are non-standard; but mining pool software relies on the scriptSig of the coinbase being arbitray.
We enforce standardness on all txs, so blocks having non-pushonly coinbases scriptSigs will be rejected, which is not what we want.

This will remove the "scriptsig-not-pushonly" restriction for standardness on coinbase inputs.

### DIFF
--- a/src/policy/policy.cpp
+++ b/src/policy/policy.cpp
@@ -88,7 +88,8 @@ bool IsStandardTx(const CTransaction &tx, bool permit_bare_multisig,
             reason = "scriptsig-size";
             return false;
         }
-        if (!txin.scriptSig.IsPushOnly()) {
+        // Coinbase can have non-pushonly scriptSig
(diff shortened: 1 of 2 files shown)


src/policy/policy.cpp
```diff
@@ -88,7 +88,7 @@ bool IsStandardTx(const CTransaction &tx, bool permit_bare_multisig,
             reason = "scriptsig-size";
             return false;
         }
-        if (!txin.scriptSig.IsPushOnly()) {
+        if (!txin.prevout.IsNull() && !txin.scriptSig.IsPushOnly()) {
```
schancel, line 91: Isn't there an IsCoinBase check that would be more accurate?

schancel (approved):

PR #151, opened 2021-04-22
New Lotus block header

EyeOfPython:
Updates the block header layout and hashing:
- Multi-layered block hashing allowing different levels of validation
- Block size for rejecting oversized blocks early on
- Block height for cheap DAA verification, and for smart contracts
- 5040 block epoch for smart contracts and sparse block header validation
- Extended metadata for adding metadata to block headers in a canonical way
- Widens the time to 6 bytes to delay the 2106 Problem by around 8M years
- Widens nonce to 8 bytes for simplified pool operations
- Shrinks block version to 1 byte and enforces versions as consensus
- Reduce the powLimit by approx. 16x to make it an even power of 2 for a later subsidy change


src/primitives/block.cpp
```diff
@@ -9,16 +9,34 @@
 #include <tinyformat.h>
 
 BlockHash CBlockHeader::GetHash() const {
-    return BlockHash(SerializeHash(*this));
+    CHashWriter layer3(SER_GETHASH, 0);
```
schancel, line 12: 💪

schancel:
Awesome!

schancel (approved):

PR #155, opened 2021-04-25
Refactor GetBlockSubsidy to depend on nBits instead of nHeight

EyeOfPython:
This will allow us to build a coin where the block reward depends on difficulty.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -860,7 +860,8 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
     return true;
 }
 
-Amount GetBlockSubsidy(int nHeight, const Consensus::Params &consensusParams) {
+Amount GetBlockSubsidy(uint32_t nBits,
+                       const Consensus::Params &consensusParams) {
     return SUBSIDY;
 }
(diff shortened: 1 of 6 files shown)


src/validation.cpp
```diff
@@ -860,7 +860,8 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
     return true;
 }
 
-Amount GetBlockSubsidy(int nHeight, const Consensus::Params &consensusParams) {
+Amount GetBlockSubsidy(uint32_t nBits,
```
schancel, line 863: What about taking the block itself? Though, it doesn't matter much I suppose. Can be changed again if we have to hardfork.

EyeOfPython: Makes it harder to test tbh. I thought about it but I'd rather keep it so it's just what we absolutely need; can refactor later if we need more, it is only used at like 3 places anyway.

schancel (approved):

src/validation.cpp
schancel: Yeah, I see this. This is good.

PR #157, opened 2021-04-26
Update default ports to be based off p2p port 10605

### LINKED ISSUE #153: Update port to 7633
Bitcoin ABC and Bitcoin Cash use the same port as Bitcoin, which prevents them from being ran on the same machine as Bitcoin Core, or each other. However, this was done due to the fact that they forked the ledger from these other networks. Because Lotus is starting fresh, we can use a different port to enable infrastructure providers to more easily spin up a our daemon. 

We should use port 7633

schancel:
Since Lotus is not a ledger fork, there is no reason to maintain the
same default ports. In fact, this makes it harder for individuals to run
nodes on the same server as whatever of flavor of bitcoind they prefer.
This we're going to move to a different port to support ease of use for
node operators.

New ports:
* mainnet
  * p2p - 10605
  * rpc - 10604
* testnet
  * p2p - 11605
  * rpc - 11604
* regtest
  * p2p - 12605
  * rpc - 12604

Closes #153

### DIFF
--- a/contrib/linearize/example-linearize.cfg
+++ b/contrib/linearize/example-linearize.cfg
@@ -5,13 +5,13 @@ rpcpassword=somepassword
 host=127.0.0.1
 
 #mainnet default
-port=8332
+port=10604
 
 #testnet default
-#port=18332
+#port=11604
 
 #regtest default
-#port=18443
+#port=12604
 
 # bootstrap.dat hashlist settings (linearize-hashes)
 max_height=313000
--- a/contrib/aur/common/bitcoin.conf
+++ b/contrib/aur/common/bitcoin.conf
@@ -1,4 +1,4 @@
-rpcport=8332
+rpcport=10604
 rpcuser=bitcoin
 rpcpassword=secret
 rpcallowip=127.0.0.1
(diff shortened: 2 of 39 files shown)


contrib/teamcity/build-configurations.py
```diff
@@ -553,6 +553,12 @@ def main():
             str(default_config_path)
         )
     )
+    parser.add_argument(
```
EyeOfPython, line 556: ?

EyeOfPython (commented):
Nothing much to review here tbh.. But can we remove the seeds in a different PR? Seems a bit out of place

contrib/teamcity/build-configurations.py
schancel: Yeah IDK. Some autoformatting happened it looks like, but why I don't know.

contrib/linearize/example-linearize.cfg
```diff
@@ -5,13 +5,13 @@ rpcpassword=somepassword
...
-port=8332
+port=10604
 
 #testnet default
-#port=18332
+#port=11604
 
 #regtest default
-#port=18443
+#port=11603
```
EyeOfPython, line 14: should be 12604

EyeOfPython (changes requested):
Regtest ports are wrong, and a few search+replace errors

EyeOfPython (approved):

PR #158, opened 2021-04-26
Change GetBlockSubsidy to emit coins based on log(difficulty)

EyeOfPython:
This dampens price volatility slightly

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -64,33 +64,32 @@ BlockAssembler MinerTestingSetup::AssemblerForTest(const CChainParams &params) {
 }
 
 constexpr uint64_t blockinfo[] = {
-    80806239248,  36663315142,  45227949834,  44108617520,  29708483491,
-    55956396199,  43373136901,  184568954235, 8030821652,   135778740536,
-    282307931074, 72805091384,  70057026190,  231652693905, 92118978074,
-    26259783141,  16287580758,  229659423653, 82899115750,  27656695552,
-    26397677886,  16393879408,  20666850084,  48483806822,  144440379235,
-    63199845025,  98834152696,  52012298551,  45492959093,  73741296661,
-    34989463980,  479411683635, 9120327637,   25020836323,  94223331090,
-    60007814754,  34427721593,  18541987412,  22714152116,  63616749685,
-    63538065054,  1438750169,   44922323154,  79332789140,  17389358740,
-    23679299233,  11045727895,  85258772653,  35271683886,  56263954376,
-    137163317583, 54269501478,  38739919713,  48034037181,  17278330193,
-    14028073987,  4849365943,   47252703650,  70772475851,  47222270790,
-    4037710411,   57543983455,  94602857448,  54810893610,  274628454954,
-    37602983315,  29815244305,  48564102352,  15841218732,  98473244519,
-    2969487671,   320310760730, 86825715486,  47504878491,  93931764317,
-    17813199340,  37203154077,  25161553257,  48611045685,  84870380488,
-    18994126776,  42600068452,  94271571206,  36882473273,  10302521845,
-    14510408253,  55153450456,  211714477708, 64276514693,  26386432305,
-    16853942118,  91499119664,  3225146899,   60966421175,  104858107378,
-    85165060220,  103395867392, 27676058438,  885215186,    882434150,
-    922772064,    28759511460,  124877325486, 86222419635,  21828564280,
-    162971421003, 62803422958,  4459801437,   56940414056,  39760054254,
+    80806239248,  65799807021,  604811529080, 126913528961, 127051102755,
+    21095037785,  47895049626,  32828337979,  22831555262,  50835325299,
+    20313599487,  158271518541, 5655489867,   11538952148,  68009355488,
+    94050716609,  30732139119,  48355079627,  14929580175,  69247195963,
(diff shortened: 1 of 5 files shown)


src/validation.cpp
```diff
@@ -862,7 +863,62 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
...
+
+    // log of max target, rounded up. It actually is 223.99997798605273, we
+    // correct for this by calculating with mantissa + 1.
+    static constexpr int64_t LOG_MAX_TARGET = 224;
+    const int32_t exp = nBits >> 24;
+    const uint32_t mantissa = (nBits & 0x007fffff);
+
+    // Compute a*log(nMantissa)
+    // We re-interpret mantissa as fixed point number with 16 bits in the
+    // fractional part, this is like dividing mantissa by 2^16.
```
schancel, line 899: > this is like dividing mantissa by 2^16

You mean multiplying?

src/validation.cpp
```diff
@@ -862,7 +863,62 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
...
+
+    // Multiply by SUBSIDY before going back from fixed point to keep the
+    // precision. Otherwise, we only increase in SUBSIDY steps.
+    const int64_t aLogMantissa =
+        (SUBSIDY_INT * logMantissaFixedPoint) >> PRECISION_BITS;
+
+    // Compute a*log(2^(8 * (exp - 3))), which can be simplified to
+    // a*(8 * (exp - 3)). We need to add PRECISION_BITS here because we
+    // implicitly divided mantissa by 2^16 before and correct for that here.
+    const int64_t aLogExponent = SUBSIDY_INT * (8 * (exp - 3) + PRECISION_BITS);
```
schancel, line 912: You've already divided out the precision above. I don't think you want to add it back here.

src/validation.cpp
```diff
@@ -862,7 +863,62 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
...
+    const int64_t aLogTarget = aLogMantissa + aLogExponent;
```
schancel, line 913: Nominally this would be:

```
aLogTarget = ((aLogMantissa << precision ) + (8 * (exp - (3 << precision)))) >> precision
```
But you already shifted the precision out above.

src/validation.cpp
```diff
@@ -862,7 +863,62 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
...
+    static constexpr int64_t PRECISION_BITS = 16;
+
+    // log of max target, rounded up. It actually is 223.99997798605273, we
+    // correct for this by calculating with mantissa + 1.
+    static constexpr int64_t LOG_MAX_TARGET = 224;
+    const int32_t exp = nBits >> 24;
+    const uint32_t mantissa = (nBits & 0x007fffff);
+
+    // Compute a*log(nMantissa)
+    // We re-interpret mantissa as fixed point number with 16 bits in the
```
schancel, line 898: We're not reinterpreting the mantissa as a fixed point, rather Log2FixedPoint gives back a number in fixed point.

src/validation.cpp
EyeOfPython: No, the above comment says it. Re-interpreting the mantissa that way basically divides the number by 2^16.

src/validation.cpp
EyeOfPython: No, we are. Log2FixedPoint expects a fixed point number, but we provide a normal integer. Log2FixedPoint will treat it as fixed point, so we correct for that later.

src/validation.cpp
EyeOfPython: No, we need to do both. mantissa gets implicitly divided by 2^16 in line 901 and the log is returned fixed point number. To convert back to normal integers, we divide by 2^16. In line 912, we then correct for the implicit division we did by reinterpreting mantissa as fixed point number.

src/validation.cpp
EyeOfPython: why the `(3 << precision)`?

src/validation.cpp
```diff
@@ -862,7 +863,62 @@ static bool WriteBlockToDisk(const CBlock &block, FlatFilePos &pos,
...
+    // fractional part, this is like dividing mantissa by 2^16.
+    // We correct for this later.
+    const int64_t logMantissaFixedPoint =
+        Log2FixedPoint(mantissa, PRECISION_BITS);
+
+    // Multiply by SUBSIDY before going back from fixed point to keep the
+    // precision. Otherwise, we only increase in SUBSIDY steps.
+    const int64_t aLogMantissa =
+        (SUBSIDY_INT * logMantissaFixedPoint) >> PRECISION_BITS;
```
schancel, line 907: Shouldn't this be:
```
const int64_t aLogMantissa  = SUBSIDY_INT * (logMantissaFixedPoint + (precision << precision)) >> precision
```

src/test/miner_tests.cpp
```diff
@@ -342,6 +343,17 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity) {
             }
 #endif
             CMutableTransaction txCoinbase(*pblock->vtx[0]);
+            // Block subsidy adjustst with difficulty, so we have to update the
```
schancel, line 346: `adjustst` -> `adjusts`

schancel (approved):

schancel (approved):

PR #162, opened 2021-04-29
Update network client name to lotusd from Bitcoin ABC

schancel:
As per title

### DIFF
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -11,7 +11,7 @@
  * for both bitcoind and bitcoin-qt, to make it harder for attackers to
  * target servers or GUI users specifically.
  */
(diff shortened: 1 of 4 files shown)


EyeOfPython:
Not "lotusd" ?

schancel:
> Not "lotusd" ?

Debatable. It wasn't bitcoind before.

EyeOfPython:
> Debatable. It wasn't bitcoind before.

It was Bitcoin ABC, which used to be the client name for Bitcoin Cash. Having a distinct name allows others to have their own node implementation.

lotusd makes sense, as that's the name of the repo.

EyeOfPython (approved):

PR #163, opened 2021-04-29
Update Bitcoin-Qt strings from Bitcoin ABC -> Lotus

schancel:
As per title

### DIFF
--- a/src/qt/bitcoinstrings.cpp
+++ b/src/qt/bitcoinstrings.cpp
@@ -7,178 +7,256 @@
 #define UNUSED
 #endif
 static const char UNUSED *bitcoin_strings[] = {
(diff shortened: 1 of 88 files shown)


EyeOfPython:
Hmmm.. Should we use the string "lotus-qt" instead maybe? I'm a little uncomfortable with using the same string for everything, makes it harder to search for in the future, what do you think?

schancel:
> Hmmm.. Should we use the string "lotus-qt" instead maybe? I'm a little uncomfortable with using the same string for everything, makes it harder to search for in the future, what do you think?

You can always search for the context which is the xml tags. The display context of these wouldn't necessarily make as much sense with "Lotus-QT".

EyeOfPython (approved):

PR #166, opened 2021-04-29
Rename binaries from bitcoin* to lotus*

schancel:
As per title.


contrib/debian/lotus-qt.lintian-overrides
```diff
@@ -1,2 +1,2 @@
 # Linked code is Expat - only Debian packaging is GPL-2+
-bitcoind: possible-gpl-code-linked-with-openssl
+lotus-qt: possible-gpl-code-linked-with-openssl
```
schancel, line 2: This is git being confused about renames. There's another file below

PR #167, opened 2021-05-01
Remove BIP9 activation logic

EyeOfPython:
We perform upgrades by doing scheduled hardforks every 6 months, so we don't need this code anymore. Also, we will remove nVersion from the block header later.


schancel (approved):
🔥

PR #170, opened 2021-05-04
Fix getblocktemplate for testnet

EyeOfPython:
This fixes a bug where when the testnet resets the nBits to 0x1c100000, the coinbase outputs would remain unchanged, resulting in invalid blocks.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -908,14 +908,20 @@ static UniValue getblocktemplate(const Config &config,
     // pointer for convenience
     CBlock *pblock = &pblocktemplate->block;
 
-    // Update nTime
+    // Update nTime (can update nBits on testnet)
     UpdateTime(pblock, chainparams, pindexPrev);
     pblock->nNonce = 0;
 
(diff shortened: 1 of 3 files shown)


src/rpc/mining.cpp
```diff
@@ -915,7 +915,11 @@ static UniValue getblocktemplate(const Config &config,
     UniValue aCaps(UniValue::VARR);
     aCaps.push_back("proposal");
 
-    Amount coinbasevalue = Amount::zero();
+    const Consensus::Params &consensusParams = chainparams.GetConsensus();
+    Amount amountFeeReward =
+        GetBlockRewardFromFees(-1 * pblocktemplate->entries[0].fees);
```
schancel, line 920: `entries[0]`? Isn't there something attached to the template itself?

schancel (approved):

src/rpc/mining.cpp
EyeOfPython: Well, we could recalculate all nFees from entries, but it's already stored in the entry of the coinbase.

Hope we can replace this with something simpler eventually.

PR #172, opened 2021-05-09
Implement raw encoding and decoding on XAddress format

schancel:
This is the first commit required to implement parsing XAddress
formats. The basic idea is that addresses are composed of 4 parts:

1. token prefix (all lowercase word characters)
2. network byte (capital word character)
3. base58 encoded payload
4. base58 encoded 4-byte weak sha256d hash of (1)|| (2) || (3)

A follow up commit will implement processing into a `CTxDestination`
to support use in the rest of the node.

### DIFF
--- a/src/addresses/xaddress.cpp
+++ b/src/addresses/xaddress.cpp
@@ -14,9 +14,9 @@
 
 namespace XAddress {
 /**
- * Create a checksum.
+ * Create SHA256 Hash of the address contents for integrity checks.
  */
-uint256 CreateCheck(const Content &addressContent) {
+uint256 HashAddressContents(const Content &addressContent) {
     CHashWriter hasher(SER_GETHASH, 0);
     hasher << addressContent.token;
     hasher << uint8_t(addressContent.network);
@@ -30,7 +30,7 @@ uint256 CreateCheck(const Content &addressContent) {
  */
 std::string Encode(const Content &addressContent) {
     std::vector<uint8_t> preencodedBuffer;
-    uint256 check = CreateCheck(addressContent);
+    uint256 check = HashAddressContents(addressContent);
     preencodedBuffer.reserve(addressContent.payload.size() + 5);
     preencodedBuffer.push_back(addressContent.type);
     preencodedBuffer.insert(preencodedBuffer.end(),
@@ -48,13 +48,13 @@ std::string Encode(const Content &addressContent) {
 /**
  * Decode a XAddress string.
  */
-bool Decode(const std::string &address, Content &parsedOutput) {
+DecodeError Decode(const std::string &address, Content &parsedOutput) {
     const auto networkPosition =
         std::find_if(address.begin(), address.end(), [](const char c) -> bool {
             return std::isupper(c) || std::isdigit(c) || c == '_';
         });
     if (networkPosition == address.end()) {
-        return false;
+        return NO_NETWORK_POSITION;
     }
     const std::string token =
         address.substr(0, std::distance(address.begin(), networkPosition));
@@ -65,20 +65,20 @@ bool Decode(const std::string &address, Content &parsedOutput) {
     // vch length cannot be greater than encodedPayload
     std::vector<uint8_t> vch;
     if (!DecodeBase58(encodedPayload, vch, encodedPayload.length())) {
-        return false;
+        return BASE58_DECODE_FAILED;
     }
     // Undersized payload. Can't include type byte and checksum
     if (vch.size() < 5) {
-        return false;
+        return UNDERSIZED_PAYLOAD;
     }
     const AddressType addressByte = AddressType(vch[0]);
     parsedOutput = Content(token, networkByte, addressByte,
                            std::vector(vch.begin() + 1, vch.end() - 4));
-    const uint256 check = CreateCheck(parsedOutput);
+    const uint256 check = HashAddressContents(parsedOutput);
     if (memcmp(&check, &vch[vch.size() - 4], 4)) {
-        return false;
+        return INTEGRITY_CHECK_FAILED;
     }
-    return true;
+    return DECODE_OK;
 }
 
 } // namespace XAddress
--- a/src/addresses/xaddress.cpp
+++ b/src/addresses/xaddress.cpp
@@ -10,89 +10,74 @@
 #include <util/string.h>
 #include <util/vector.h>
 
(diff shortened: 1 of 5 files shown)


src/addresses/xaddress.cpp
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017 Pieter Wuille
+// Copyright (c) 2017-2019 The Bitcoin developers
```
EyeOfPython, line 2: no

src/addresses/xaddress.h
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2017 Pieter Wuille
+// Copyright (c) 2017-2018 The Bitcoin developers
```
EyeOfPython, line 2: no

src/base58.cpp
```diff
@@ -14,6 +14,8 @@
 #include <cstring>
 #include <limits>
 
+#include <iostream>
```
EyeOfPython, line 17: ?

src/test/xaddress_tests.cpp
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2011-2019 The Bitcoin Core developers
```
EyeOfPython, line 1: no

EyeOfPython (changes requested):

src/addresses/xaddress.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <uint256.h>
+#include <util/strencodings.h>
+#include <util/string.h>
+#include <util/vector.h>
+
+#include <string>
+
+namespace XAddress {
+/**
+ * Create a checksum.
```
EyeOfPython, line 17: This is quite confusing. Can you make this return just the 4 bytes?

Also rename to something like "CalcChecksum"

EyeOfPython (changes requested):

src/addresses/xaddress.cpp
schancel: Doing that would just involve another copy, but I can do it if you want. Renaming it to that would be incorrect. It's not a checksum, and the existing Base58 library uses the term `Check`

EyeOfPython: Why is it not a checksum? Because it's not reversible?

src/test/xaddress_tests.cpp
```diff
@@ -0,0 +1,82 @@
...
+const std::string pszBase58 =
+    "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+
+BOOST_FIXTURE_TEST_SUITE(xaddress_tests, BasicTestingSetup)
+
+BOOST_AUTO_TEST_CASE(raw_encode_decode_succeeds) {
+    // First create an instance of an engine.
+    std::random_device rnd_device;
+    // Specify the engine and distribution.
+    std::mt19937 mersenne_engine{rnd_device()}; // Generates random integers
```
EyeOfPython, line 27: We can just use the `MMIXLinearCongruentialGenerator` in `<lgc.h>`; tests don't have to be cryptographically secure

EyeOfPython (changes requested):

src/addresses/xaddress.h
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Encode an XAddress string.
+ */
+std::string Encode(const Content &addressContent);
+/**
+ * Decode an XAddress string. Returns false on failure, and parsedOutput is not
+ * modified.
+ */
+bool Decode(const std::string &address, Content &parsedOutput);
```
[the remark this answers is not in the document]
schancel: Possibly, but this is an internal function. Generally the one that returns a CTxDestination will be used in the codebase right now, and that only returns a `CNoDestination` on error right now. I need to be able to use this alongside those existing APIs for the time being.

src/addresses/xaddress.h
```diff
@@ -0,0 +1,50 @@
...
+    REGTEST = 'R',
+};
+
+enum AddressType : uint8_t {
+    SCRIPT_PUB_KEY = 0,
+};
+
+class Content {
+public:
+    std::string token;
```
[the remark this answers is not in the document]
schancel: 🤷 I can change it, but even that isn't used consistently.

src/addresses/xaddress.cpp
schancel: Generally this can't tell you where the error is.

schancel: Either way, I need to fix the comment.

src/test/xaddress_tests.cpp
schancel: Sure

src/addresses/xaddress.cpp
EyeOfPython: Ok, so how do we call that part then? Just `check`? I don't like that; it implies something different, and is used in other places as "the act of checking".

According to [Wikipedia](https://en.wikipedia.org/wiki/Checksum), a checksum doesn't require to be able to tell where an error is: "The simplest checksum algorithm is the so-called longitudinal parity check", which doesn't tell you where the error is.

I'd prefer if we used the term "checksum", pretty much everyone will know what it means this way.

src/addresses/xaddress.h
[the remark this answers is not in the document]
EyeOfPython: It isn't, but I think it's a good convention.

src/addresses/xaddress.h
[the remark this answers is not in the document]
EyeOfPython: Ok, sounds good. In Rust, this would have been `-> Result<Address, AddressError>`.

We can keep it as-is.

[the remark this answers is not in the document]
schancel: Yeah. We could do that in C++ too eventually. I am a fan of returning an `std::pair` and using `std::tie` on it: cashaddr.cpp and cashaddrenc.cpp do this -- but not for errors.

src/test/xaddress_tests.cpp
schancel: Actually, that doesn't generate uniform data, right? This isn't trying to span some range, it's just a random data payload. What's wrong with this? This already isn't cryptographically secure anyways.

EyeOfPython (approved):

PR #176, opened 2021-05-15
New Lotus sighash

EyeOfPython:
This replaces (but not yet renames) the BIP341 sighash introduced two months ago.

It allows smart contracts to more easily access parts of the transaction, e.g. by hashing the inputs/outputs in a merkle tree.


schancel:
Let's just call it SIGHASH_RHIZOME or SIGHASH_LOTUS_ROOT or something

src/primitives/transaction.h
```diff
@@ -331,12 +331,22 @@ class CCoinsView;
...
     std::vector<CTxOut> m_spent_outputs;
-    uint256 m_prevouts_single_hash;
-    uint256 m_sequences_single_hash;
-    uint256 m_outputs_single_hash;
-    uint256 m_spent_amounts_single_hash;
-    uint256 m_spent_scripts_single_hash;
+    /** Merkle root of the H(input.prevout || input.nSequence) for all inputs */
+    uint256 m_inputs_merkle_root;
+    /** Merkle root of the coins spend by the tx */
+    uint256 m_inputs_spent_outputs_merkle_root;
```
schancel, line 339: 👍 I was thinking we want to make sure not to rehash all these if someone specifies these flags.

src/script/interpreter.cpp
```diff
@@ -1734,21 +1731,21 @@ bool SignatureHashBIP143(uint256 &sighashOut, const CScript &scriptCode,
     uint256 hashOutputs;
 
     if (!sigHashType.hasAnyoneCanPay()) {
-        hashPrevouts = cache ? cache->hashPrevouts
-                             : SHA256Uint256(GetPrevoutsSHA256(txTo));
+        hashPrevouts =
+            cache ? cache->hashPrevouts : GetBIP143PrevoutsHash(txTo.vin);
```
schancel, line 1735: Does this sometimes not have the cache as compared with the BIP341 sigs?

src/script/interpreter.cpp
```diff
@@ -1613,68 +1614,64 @@ template <class T> class CTransactionSignatureSerializer {
...
+    hashOutputs = GetBIP143OutputsHash(txTo.vout);
     m_spent_outputs = std::move(spent_outputs);
-    m_spent_amounts_single_hash = GetSpentAmountsSHA256(m_spent_outputs);
-    m_spent_scripts_single_hash = GetSpentScriptsSHA256(m_spent_outputs);
+    if (m_spent_outputs.size() > 0) {
+        m_inputs_merkle_root =
+            TxInputsMerkleRoot(txTo.vin, m_inputs_merkle_height);
+        m_outputs_merkle_root =
+            TxOutputsMerkleRoot(txTo.vout, m_outputs_merkle_height);
+        m_amount_outputs_sum = Amount::zero();
```
schancel, line 1658: Shouldn't this always be initialized?

src/script/interpreter.cpp
```diff
@@ -1613,68 +1614,64 @@ template <class T> class CTransactionSignatureSerializer {
...
+        for (const CTxOut &output : txTo.vout) {
+            m_amount_outputs_sum += output.nValue;
+        }
+        {
+            m_amount_inputs_sum = Amount::zero();
```
schancel, line 1663: Why is this under the m_spent outs check size check?

src/script/interpreter.cpp
```diff
@@ -1780,82 +1777,70 @@ bool SignatureHashBIP143(uint256 &sighashOut, const CScript &scriptCode,
...
     if (!(hash_type & 0x03) || (hash_type & 0x1c)) {
-        // hash_type 0 is invalid, any undefined bits are invalid 
+        // hash_type 0 is invalid, any undefined bits are invalid
         return false;
     }
     ss << hash_type;
 
-    // Transaction level data
+    {
+        const uint8_t spend_type = ext_flag << 1;
```
schancel, line 1802: What's this shift for exactly?

src/script/interpreter.cpp
```diff
@@ -1780,82 +1777,70 @@ bool SignatureHashBIP143(uint256 &sighashOut, const CScript &scriptCode,
...
+        input_hash << tx_to.vin[in_pos].prevout;
+        input_hash << tx_to.vin[in_pos].nSequence;
+        input_hash << cache.m_spent_outputs[in_pos];
+        ss << input_hash.GetHash();
+    }
+
+    // Additional data for non-taproot signatures (see BIP342)
+    if (execdata) {
+        ss << execdata->m_codeseparator_pos;
+        ss << execdata->m_executed_script_hash;
```
schancel, line 1814: This still has the tagged hash, right?

src/script/interpreter.cpp
```diff
@@ -1780,82 +1777,70 @@ bool SignatureHashBIP143(uint256 &sighashOut, const CScript &scriptCode,
...
-        ss << in_pos;
+    ss << tx_to.nVersion;
+    if (!sig_hash_type.hasAnyoneCanPay()) {
+        ss << cache.m_inputs_merkle_root;
+        ss << uint8_t(cache.m_inputs_merkle_height);
     }
-
-    // Data about the output (if only one).
     if (sig_hash_type.getBaseType() == BaseSigHashType::SINGLE) {
         if (in_pos >= tx_to.vout.size()) {
```
schancel, line 1832: I don't understand this check in relation to SIGHASH::SINGLE

src/script/interpreter.cpp
```diff
@@ -1613,68 +1614,64 @@ template <class T> class CTransactionSignatureSerializer {
...
-    hashPrevouts = SHA256Uint256(m_prevouts_single_hash);
-    hashSequence = SHA256Uint256(m_sequences_single_hash);
-    hashOutputs = SHA256Uint256(m_outputs_single_hash);
+    hashPrevouts = GetBIP143PrevoutsHash(txTo.vin);
+    hashSequence = GetBIP143SequencesHash(txTo.vin);
+    hashOutputs = GetBIP143OutputsHash(txTo.vout);
     m_spent_outputs = std::move(spent_outputs);
-    m_spent_amounts_single_hash = GetSpentAmountsSHA256(m_spent_outputs);
-    m_spent_scripts_single_hash = GetSpentScriptsSHA256(m_spent_outputs);
+    if (m_spent_outputs.size() > 0) {
```
schancel, line 1653: This is an error, correct? So why not return here?

src/script/interpreter.cpp
```diff
@@ -1613,68 +1614,64 @@ template <class T> class CTransactionSignatureSerializer {
...
+        m_outputs_merkle_root =
+            TxOutputsMerkleRoot(txTo.vout, m_outputs_merkle_height);
+        m_amount_outputs_sum = Amount::zero();
+        for (const CTxOut &output : txTo.vout) {
+            m_amount_outputs_sum += output.nValue;
+        }
+        {
+            m_amount_inputs_sum = Amount::zero();
+            std::vector<uint256> spent_outputs_hashes;
+            spent_outputs_hashes.reserve(txTo.vin.size());
```
schancel, line 1665: Why shouldn't this be m_spent_outputs.size()? You're copying them. If there is an invariant that these two sizes need to be the same, then that should be asserted instead of implied.

src/script/interpreter.cpp
```diff
@@ -1613,68 +1614,64 @@ template <class T> class CTransactionSignatureSerializer {
...
+            for (const CTxOut &spent_output : m_spent_outputs) {
+                spent_outputs_hashes.push_back(SerializeHash(spent_output));
+                m_amount_inputs_sum += spent_output.nValue;
+            }
+            size_t dummy;
+            m_inputs_spent_outputs_merkle_root =
```
schancel, line 1671: We need this height for the same reason as with the txid.

EyeOfPython: I can add an assert

src/script/interpreter.cpp
EyeOfPython: Core

schancel (approved):

PR #177, opened 2021-05-16
Revert mistaken changes to gitian build yaml configs

schancel:
As per title. In the zeal to rename everything to lotus, some changes
to the gitian linux and windows scripts were erroneously included. This
commit reverts those changes. Longer term, more references to bitcoin
should be swapped to lotus to avoid confusion, but within these scripts
the references are immaterial.


EyeOfPython (approved):
LGTM

PR #179, opened 2021-05-17
Implement Lotus XAddresses throughout code base

schancel:
In previous commits we implemented parsing and encoding for Lotus
XAddresses. This commit implements them throughout the QT GUI, the RPC,
and lotus-tx as the default address format. CashAddrs and Legacy
addresses are still supported, but are automatic fallbacks.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -2798,9 +2798,9 @@ bool AppInitMain(Config &config, RPCServer &rpcServer,
         return false;
     }
 
-    // Encoded addresses using cashaddr instead of base58.
-    // We do this by default to avoid confusion with BTC addresses.
-    config.SetCashAddrEncoding(args.GetBoolArg("-usecashaddr", true));
+    // By default, we use XAddress; this option will make it use the CashAddr
+    // format.
+    config.SetCashAddrEncoding(args.GetBoolArg("-usecashaddr", false));
 
     // Step 8: load indexers
     if (args.GetBoolArg("-txindex", DEFAULT_TXINDEX)) {
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -4,7 +4,8 @@
 
 #include <qt/guiutil.h>
 
-#include <cashaddrenc.h>
+#include <addresses/xaddress.h>
+#include <cashaddr.h>
 #include <chainparams.h>
 #include <interfaces/node.h>
 #include <key_io.h>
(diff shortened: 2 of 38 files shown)


src/init.cpp
```diff
@@ -2798,9 +2798,8 @@ bool AppInitMain(Config &config, RPCServer &rpcServer,
         return false;
     }
 
-    // Encoded addresses using cashaddr instead of base58.
-    // We do this by default to avoid confusion with BTC addresses.
-    config.SetCashAddrEncoding(args.GetBoolArg("-usecashaddr", true));
+    // Encode addresses using XAddress by default instead of CashAddr.
```
EyeOfPython, line 2801: I don't like this comment.

Should be: "By default, we use XAddress; this option will make it use the CashAddr format"

src/qt/guiutil.cpp
```diff
@@ -129,15 +129,26 @@ void setupAddressWidget(QValidatedLineEdit *widget, QWidget *parent) {
...
+    const std::string addr = (uri.scheme() + ":" + uri.path()).toStdString();
+    auto decoded = cashaddr::Decode(addr, "");
+    return !decoded.first.empty();
+}
+
 bool parseBitcoinURI(const QString &scheme, const QUrl &uri,
                      SendCoinsRecipient *out) {
     // return if URI has wrong scheme.
-    if (!uri.isValid() || uri.scheme() != scheme) {
+    if (!uri.isValid() || !(uri.scheme() == scheme || uri.scheme() == "payto" || uri.scheme() == "")) {
```
EyeOfPython, line 141: Why does the "payto" here have to be hard-coded?

EyeOfPython (changes requested):
Looks mostly good. Just two points

src/qt/guiutil.cpp
schancel: What should it be?

EyeOfPython: I was thinking a constant somewhere

EyeOfPython: I was thinking a constant somewhere

schancel: I added a constant in the header file.

src/qt/test/guiutiltests.cpp
```diff
@@ -9,6 +9,8 @@
 #include <key_io.h>
 #include <qt/guiutil.h>
 
+#include <iostream>
```
EyeOfPython, line 12: remove

src/qt/test/uritests.cpp
```diff
@@ -12,6 +12,94 @@
 
 #include <QUrl>
 
+#include <iostream>
```
EyeOfPython, line 15: remove

EyeOfPython (changes requested):
Looks almost ready to go.

EyeOfPython (approved):
I don't feel too good on the `char ADDRESS_SCHEME[]`, but I'll just trust you that it really is necessary as the comment implies.

schancel:
> I don't feel too good on the `char ADDRESS_SCHEME[]`, but I'll just trust you that it really is necessary as the comment implies.

Just really no point in having to convert it from a std::string to a QString. It gets automatically converted if it's left as a C-string array in both cases.

PR #181, opened 2021-05-22
Implement random selection of output addresses based on epoch block

schancel:
We've decided to implement a lottocracy for the system of governance on
Lotus. As such, each weak (epoch) a set of random citizens will be
selected to receive a portion of the block reward which they may
allocate however they want. This code implements the random selection
from a list of addresses of verified citizens. This is subject to modulo
bias, however the effect is miner given the use case. In the future,
better algorithms may be implemented.

### DIFF
--- a/src/minerfund.cpp
+++ b/src/minerfund.cpp
@@ -17,11 +17,18 @@ static const CTxOut BuildOutput(const std::string &dest, const Amount amount) {
     return {amount, script};
 }
 
-static const CTxOut BurnOutput(const Amount amount) {
-    std::vector<uint8_t> burnPayload = {
-        200, 225, 246, 229, 160, 237, 229, 242, 227, 249, 160, 239, 238,
-        160, 237, 229, 172, 160, 225, 160, 243, 233, 238, 238, 229, 242};
-    return {amount, CScript() << OP_RETURN << burnPayload};
+const CTxOut BuildRandomOutput(const std::vector<std::string> &addressList,
+                               uint64_t slot, Amount shareAmount,
+                               uint256 epochBlockHash) {
+    CHashWriter hasher(SER_GETHASH, 0);
+    hasher << slot;
+    hasher << epochBlockHash;
+    uint256 hash = hasher.GetSHA256();
+    uint64_t weakHash = hash.GetUint64(0);
+    // Note: There is modulo bias here, but it is being ignored as the random
+    // domain is much much larger than injected range. Thus, the effect is
+    // small.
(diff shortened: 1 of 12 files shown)


src/minerfund.cpp
```diff
@@ -24,15 +24,30 @@ static const CTxOut BurnOutput(const Amount amount) {
     return {amount, CScript() << OP_RETURN << burnPayload};
 }
 
+std::vector<std::string> citizenAddresses = {
+    "pqnqv9lt7e5vjyp0w88zf2af0l92l8rxdgnlxww9j9",
+    "qqrad726sy24klmsd4dus5gxl0rhgz0rdcjnagpmve",
+    "qzafzyamh8rgszacw73gg2vly8gnxe099qx6fqrdfd",
+};
+
+static CTxOut BuildWinnerOutput(uint64_t slot, Amount shareAmount, uint256 seed) {
```
EyeOfPython, line 33: Should we call it `epochBlockHash` instead of `seed`?

I have to admit I'm quite delighted that we already found a really good use for the epoch hash. Imagine how much more difficult the implementation of this would be like without it just in the header.

And certainly more uses to come.

EyeOfPython: Also, if possible, we shouldn't use jargon like "winner". It should be more like "selected" or "chosen" or "representative". This is not a lottery nor raffle, so we shouldn't use terms that could confuse regulators.

src/minerfund.cpp
```diff
@@ -41,6 +56,15 @@ std::vector<CTxOut> GetMinerFundRequiredOutputs(const Consensus::Params &params,
         // Tobias
         BuildOutput("qzafzyamh8rgszacw73gg2vly8gnxe099qx6fqrdfd", shareAmount),
         // Burn the remaining shares
-        BurnOutput(10 * shareAmount),
+        BuildWinnerOutput(1, shareAmount, epochBlockHash),
```
EyeOfPython, line 59: For the testnet, can you burn 2 outputs and make the remaining 8 random? This will be more similar to the behavior of the mainnet.

EyeOfPython (changes requested):

src/minerfund.cpp
schancel: Sure

src/minerfund.cpp
schancel: ?

EyeOfPython: Mainnet will have 5 fixed and 8 random outputs. Atm, we only have 3 fixed outputs. Let's burn two and make 8 random.

Right now it's 3 fixed and 10 random.

PR #182, opened 2021-05-22
Update unit formating in QT to be 10^8 and Lotus properly

schancel:
The QT Wallet uses completely different formatters from the rest of
the codebase and was missed during the previous re-denomination patch.
This commit changes the QT wallet to use the 10^6 base unit, and ensures
the QT wallet uses the appropriate formats.

### DIFF
--- a/src/qt/bitcoinunits.cpp
+++ b/src/qt/bitcoinunits.cpp
@@ -13,18 +13,16 @@ BitcoinUnits::BitcoinUnits(QObject *parent)
 
 QList<BitcoinUnits::Unit> BitcoinUnits::availableUnits() {
     QList<BitcoinUnits::Unit> unitlist;
-    unitlist.append(BCH);
-    unitlist.append(mBCH);
-    unitlist.append(uBCH);
+    unitlist.append(LOTUS);
+    unitlist.append(mLOTUS);
     unitlist.append(SAT);
     return unitlist;
 }
 
 bool BitcoinUnits::valid(int unit) {
     switch (unit) {
-        case BCH:
-        case mBCH:
-        case uBCH:
+        case LOTUS:
+        case mLOTUS:
         case SAT:
             return true;
         default:
@@ -34,12 +32,10 @@ bool BitcoinUnits::valid(int unit) {
 
 QString BitcoinUnits::longName(int unit) {
     switch (unit) {
-        case BCH:
+        case LOTUS:
             return QString(CURRENCY_UNIT.c_str());
-        case mBCH:
+        case mLOTUS:
             return QString("m") + QString(CURRENCY_UNIT.c_str());
-        case uBCH:
-            return QString::fromUtf8("μ") + QString(CURRENCY_UNIT.c_str());
         case SAT:
             return QString("Satoshi (sat)");
         default:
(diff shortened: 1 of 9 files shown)


src/qt/bitcoinunits.cpp
```diff
@@ -58,29 +54,24 @@ QString BitcoinUnits::shortName(int unit) {
...
-        case uBCH:
-            return QString("Micro-Bitcoins (1 / 1" THIN_SP_UTF8
-                           "000" THIN_SP_UTF8 "000)");
+        case LOTUS:
+            return QString("Lotus");
+        case mLOTUS:
+            return QString("Milli-Lotus (1 / 1" THIN_SP_UTF8 "000)");
         case SAT:
             return QString("Satoshi (sat) (1 / 100" THIN_SP_UTF8
-                           "000" THIN_SP_UTF8 "000)");
```
schancel, line 70: This needs to be fixed.

src/qt/test/uritests.cpp
```diff
@@ -40,23 +40,23 @@ void URITests::uriTestsCashAddr() {
     QVERIFY(rv.amount == Amount::zero());
 
     uri.setUrl(QString(
-        "bitcoincash:qpm2qsznhks23z7629mms6s4cwef74vcwvy22gdx6a?amount=0.001"));
+        "bitcoincash:qpm2qsznhks23z7629mms6s4cwef74vcwvy22gdx6a?amount=100000"));
```
EyeOfPython, line 43:
```suggestion
        "bitcoincash:qpm2qsznhks23z7629mms6s4cwef74vcwvy22gdx6a?amount=0.1"));
```

no?

src/qt/guiutil.cpp
```diff
@@ -164,7 +164,7 @@ bool parseBitcoinURI(const QString &scheme, const QUrl &uri,
             fShouldReturnFalse = false;
         } else if (i->first == "amount") {
             if (!i->second.isEmpty()) {
-                if (!BitcoinUnits::parse(BitcoinUnits::BCH, i->second,
+                if (!BitcoinUnits::parse(BitcoinUnits::SAT, i->second,
```
EyeOfPython, line 167:
```suggestion
                if (!BitcoinUnits::parse(BitcoinUnits::LOTUS, i->second,
```
???

schancel: Per the current spec, query parameters must use sats. We can change it if you want, but it results in confusion because of differences between wallets.

src/qt/test/uritests.cpp
schancel: Has to do with the other change to the address format spec.

src/qt/guiutil.cpp
EyeOfPython: so the current Bitcoin-QT impl is wrong?

schancel: Current Bitcoin-QT implementation is correct given the lack of any specification of how the amount field should be handled within the current two address specifications. For Lotus, I'm specifying that it should always be sats. We could detect a decimal point and toggle between the two. That's what I had to do in Cashew -- but I really dislike it.

EyeOfPython: I see.

EyeOfPython (approved):

PR #184, opened 2021-05-22
Add `enableMinerFund` to config to enable testing

schancel:
This commit adds `enableMinerFund` to the config so that it can be
overriden in w/o setting gArgs. It also adds flags to the
`BlockAssembler` options as well as the validation options in order to
support passing these configs down to the appropriate code. However,
current testing setup relies on the global `GetConfig()` singleton. This
singleton always returns the chainparams from the global `Params()`
singleton function. The testing setup changes the Params after
initialization, thus we need to ensure that `EnableMinerFund()` returns
the appropriate default consensus parameter if it has not been formally
set by the command line or the testing setup. In the future, global
singletons should be removed so that tests can be ran in parallel.
However, this requires substantial refactors to the entire codebase to
ensure proper initialization ordering and encapsulation of
functionality.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -8,7 +8,8 @@
 #include <consensus/consensus.h> // DEFAULT_MAX_BLOCK_SIZE
 
 GlobalConfig::GlobalConfig()
-    : useCashAddr(false), nMaxBlockSize(DEFAULT_MAX_BLOCK_SIZE) {}
+    : useCashAddr(false), enableMinerFund(MinerFundStatus::Unset),
+      nMaxBlockSize(DEFAULT_MAX_BLOCK_SIZE) {}
 
 bool GlobalConfig::SetMaxBlockSize(uint64_t maxBlockSize) {
     // Do not allow maxBlockSize to be set below historic 1MB limit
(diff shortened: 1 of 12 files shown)


src/config.cpp
```diff
@@ -42,23 +43,41 @@ bool GlobalConfig::UseCashAddrEncoding() const {
...
 DummyConfig::DummyConfig(std::unique_ptr<CChainParams> chainParamsIn)
-    : chainParams(std::move(chainParamsIn)) {}
+    : chainParams(std::move(chainParamsIn)),
+      enableMinerFund(chainParams->GetConsensus().enableMinerFund) {}
 
 void DummyConfig::SetChainParams(std::string net) {
     chainParams = CreateChainParams(net);
 }
 
-void GlobalConfig::SetExcessUTXOCharge(Amount fee) {
```
EyeOfPython, line 58: Why is this being removed?

schancel: Should not have been, but it worked apparently because Amaury ripped out all the other code using it.

EyeOfPython (approved):

PR #185, opened 2021-05-22
Implement a new simplified mining API `getunsolvedblock`

schancel:
This API takes an address and returns a target, and the serialized block
as hex. Because the nonce was expanded, there is no need to grind on the
coinbase in order to produce a valid block. Thus, we can implement a
miner which takes only the block header. However, currently we also
include the transactions as there is no internal caching for the
transaction set. These are needed in submit block for validating the
merkle root and broadcasting the transactions later. Mining software
will need to hold on to them, but does not need to do anything with
them.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -11,6 +11,7 @@
 #include <config.h>
 #include <consensus/activation.h>
 #include <consensus/consensus.h>
+#include <consensus/merkle.h>
 #include <consensus/params.h>
 #include <consensus/validation.h>
 #include <core_io.h>
@@ -1006,6 +1007,88 @@ static UniValue getblocktemplate(const Config &config,
     return result;
 }
 
+static UniValue getrawunsolvedblock(const Config &config,
+                                    const JSONRPCRequest &request) {
+    RPCHelpMan{
+        "getrawunsolvedblock",
+        "Get a serialized block candidate with coinbase sending to the "
+        "specified address",
+        {
+            {"address", RPCArg::Type::STR, RPCArg::Optional::NO,
+             "The address to send the newly generated coins to."},
+        },
+        RPCResult{
+            RPCResult::Type::STR_HEX,
+            "blockhex",
+            "Serialized block as hex.",
+        },
+        RPCExamples{"\nCreate an unsolved block to to myaddress\n" +
+                    HelpExampleCli("getrawunsolvedblock", "\"myaddress\"") +
+                    "If you are running the Lotus wallet, you can get a new "
+                    "address to send the newly generated coins to with:\n" +
+                    HelpExampleCli("getnewaddress", "")},
+    }
+        .Check(request);
+
+    CTxDestination destination =
+        DecodeDestination(request.params[0].get_str(), config.GetChainParams());
+    if (!IsValidDestination(destination)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY,
+                           "Error: Invalid address");
+    }
+
+    const CChainParams &chainparams = config.GetChainParams();
+
+    NodeContext &node = EnsureNodeContext(request.context);
+    if (!node.connman) {
+        throw JSONRPCError(
+            RPC_CLIENT_P2P_DISABLED,
+            "Error: Peer-to-peer functionality missing or disabled");
+    }
+
+    if (node.connman->GetNodeCount(CConnman::CONNECTIONS_ALL) == 0) {
+        throw JSONRPCError(RPC_CLIENT_NOT_CONNECTED, "Node is not connected!");
+    }
+
+    if (::ChainstateActive().IsInitialBlockDownload()) {
+        throw JSONRPCError(RPC_CLIENT_IN_INITIAL_DOWNLOAD, PACKAGE_NAME
+                           " is in initial sync and waiting for blocks...");
+    }
+
+    LOCK(cs_main);
+    const CTxMemPool &mempool = EnsureMemPool(request.context);
+
+    // Store the pindexBest used before CreateNewBlock, to avoid races
+    CBlockIndex *pindexPrevNew = ::ChainActive().Tip();
+
+    // Create new block with provided coinbase output script.
+    CScript coinbase_script = GetScriptForDestination(destination);
+    std::unique_ptr<CBlockTemplate> pblocktemplate =
+        BlockAssembler(config, mempool).CreateNewBlock(coinbase_script);
+    if (!pblocktemplate) {
+        throw JSONRPCError(RPC_OUT_OF_MEMORY, "Out of memory");
+    }
+
+    // Need to update only after we know CreateNewBlock succeeded
+    CHECK_NONFATAL(pindexPrevNew);
+
+    // pointer for convenience
+    CBlock *pblock = &pblocktemplate->block;
+    // Update nTime (can update nBits on testnet)
+    UpdateTime(pblock, chainparams, pindexPrevNew);
+    pblock->nNonce = 0;
+    // Update merkle root so mining software need not touch it.
+    pblock->hashMerkleRoot = BlockMerkleRoot(*pblock);
+
+    CDataStream ssBlock(SER_NETWORK, PROTOCOL_VERSION);
+    ssBlock << *pblock;
+    arith_uint256 hashTarget = arith_uint256().SetCompact(pblock->nBits);
+    UniValue result(UniValue::VOBJ);
+    result.pushKV("blockhex", HexStr(ssBlock));
+    result.pushKV("target", hashTarget.GetHex());
+    return result;
+}
+
 class submitblock_StateCatcher final : public CValidationInterface {
 public:
     uint256 hash;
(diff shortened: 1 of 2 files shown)


test/functional/logos_rpc_getunsolvedblock.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+# Copyright (c) 2020 The Bitcoin developers
```
EyeOfPython, line 2: Fix

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
     return result;
 }
 
+static UniValue getunsolvedblock(const Config &config,
+                                 const JSONRPCRequest &request) {
+    RPCHelpMan{
+        "getunsolvedblock",
+        "Get a block candidate header with coinbase sending to specified "
+        "address.",
```
EyeOfPython, line 1014: Change to something like:
```suggestion
        "Get a serialized block candidate with coinbase sending to the specified address.",
```

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
...
+        {
+            {"address", RPCArg::Type::STR, RPCArg::Optional::NO,
+             "The address to send the newly generated bitcoin to."},
```
EyeOfPython, line 1017: bitcoin -> coins

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
...
+        RPCResult{
+            RPCResult::Type::STR_HEX,
+            "blockhex",
+            "Serialized block as hex.",
+        },
+        RPCExamples{"\nCreate an unsolved block to to myaddress\n" +
+                    HelpExampleCli("getunsolvedblock", "\"myaddress\"") +
+                    "If you are running the Lotus wallet, you can get a new "
+                    "address to send the newly generated bitcoin to with:\n" +
+                    HelpExampleCli("getnewaddress", "")},
```
EyeOfPython, line 1028: This whole text mixes Bitcoin and Lotus

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
...
+    NodeContext &node = EnsureNodeContext(request.context);
+    if (!node.connman) {
+        throw JSONRPCError(
+            RPC_CLIENT_P2P_DISABLED,
+            "Error: Peer-to-peer functionality missing or disabled");
+    }
+
+    if (node.connman->GetNodeCount(CConnman::CONNECTIONS_ALL) == 0) {
+        throw JSONRPCError(RPC_CLIENT_NOT_CONNECTED,
+                           "Bitcoin is not connected!");
```
EyeOfPython, line 1050: something like
```suggestion
                           "Node is not connected!");
```

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
...
+    }
+
+    LOCK(cs_main);
+    const CTxMemPool &mempool = EnsureMemPool(request.context);
+
+    // Store the pindexBest used before CreateNewBlock, to avoid races
+    CBlockIndex *pindexPrevNew = ::ChainActive().Tip();
+
+    // Create new block. scriptDummy is just for testing block validity and
+    // is not part of the returned template.
```
EyeOfPython, line 1065:
```suggestion
    // Create new block with provided coinbase output script
```

src/rpc/mining.cpp
```diff
@@ -1005,6 +1006,90 @@ static UniValue getblocktemplate(const Config &config,
...
+    UpdateTime(pblock, chainparams, pindexPrevNew);
+    pblock->nNonce = 0;
+    // Update merkle root so mining software need not touch it.
+    pblock->hashMerkleRoot = BlockMerkleRoot(*pblock);
+
+    CDataStream ssBlock(SER_NETWORK, PROTOCOL_VERSION);
+    ssBlock << *pblock;
+    arith_uint256 hashTarget = arith_uint256().SetCompact(pblock->nBits);
+    UniValue result(UniValue::VOBJ);
+    result.pushKV("block", HexStr(ssBlock));
```
EyeOfPython, line 1088: maybe this is more clear?

```suggestion
    result.pushKV("rawblock", HexStr(ssBlock));
```

or

```suggestion
    result.pushKV("blockhex", HexStr(ssBlock));
```

I'm not sure how other parts of the RPC name it

src/rpc/client.cpp
```diff
@@ -59,6 +59,7 @@ static const CRPCConvertParam vRPCConvertParams[] = {
     {"listtransactions", 3, "include_watchonly"},
     {"walletpassphrase", 1, "timeout"},
     {"getblocktemplate", 0, "template_request"},
+    {"getunsolvedblock", 0, "address"},
```
EyeOfPython, line 62: Shouldn't it be 1? I'm not sure what that number does.

test/functional/logos_rpc_getunsolvedblock.py
```diff
@@ -0,0 +1,49 @@
...
+        address = node.get_deterministic_priv_key().address
+        # Generate first block after genesis
+        node.generatetoaddress(1, address)
+        response = node.getunsolvedblock(address)
+        block = FromHex(CBlock(), response['block'])
+        assert_equal(block.hashMerkleRoot, block.calc_merkle_root())
+        block.solve()
+        node.submitblock(block.serialize().hex())
+        response = node.getbestblockhash()
+        assert_equal(response, format(block.sha256, 'x'))
```
EyeOfPython, line 45:
```suggestion
        assert_equal(response, block.hash)
```

test/functional/logos_rpc_getunsolvedblock.py
```diff
@@ -0,0 +1,49 @@
...
+            '-allownonstdtxnconsensus=1',
+        ], [
+            '-allownonstdtxnconsensus=1',
+        ]]
+
+    def run_test(self):
+        node = self.nodes[0]
+        address = node.get_deterministic_priv_key().address
+        # Generate first block after genesis
+        node.generatetoaddress(1, address)
```
EyeOfPython, line 38: Maybe add some dummy txs and check if they are in the decoded `block` ?

EyeOfPython (changes requested):
Lots of nits

src/rpc/client.cpp
schancel: Parameter index

schancel: Actually, this shouldn't be there.

schancel: It has to do with parameter conversion to non-strings. That explains why I was having weird JSON encoding errors w/o manually adding quotes.

test/functional/logos_rpc_getrawunsolvedblock.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Logos Foundation
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""
+Tests for Bitcoin ABC mining RPCs
```
EyeOfPython, line 6: fix

test/functional/logos_rpc_getrawunsolvedblock.py
```diff
@@ -0,0 +1,49 @@
...
+class LotusGetRawUnsolvedBlock(BitcoinTestFramework):
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 2
+        self.extra_args = [[
+            '-enableminerfund',
+            '-allownonstdtxnconsensus=1',
+        ], [
+            '-allownonstdtxnconsensus=1',
+        ]]
```
EyeOfPython, line 32: Seems a bit excessive.

test/functional/logos_rpc_getrawunsolvedblock.py
```diff
@@ -0,0 +1,65 @@
...
+
+NUM_TRANSACTIONS = 10
+
+pp = pprint.PrettyPrinter(depth=6)
+
+
+class LotusGetRawUnsolvedBlock(BitcoinTestFramework):
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 2
```
EyeOfPython, line 31: Just remove this line.

Also, isn't it simpler with a setup chain? I think you can do "sendtoaddress" right away IIRC.

test/functional/logos_rpc_getrawunsolvedblock.py
```diff
@@ -0,0 +1,65 @@
...
+    FromHex,
+)
+
+from decimal import Decimal
+
+import pprint
+
+NUM_TRANSACTIONS = 10
+
+pp = pprint.PrettyPrinter(depth=6)
```
EyeOfPython, line 25: small pp

EyeOfPython (changes requested):

test/functional/logos_rpc_getrawunsolvedblock.py
schancel: You can't mine without a second node.

test/functional/logos_rpc_getrawunsolvedblock.py
schancel: gar

test/functional/logos_rpc_getrawunsolvedblock.py
schancel: Just checked, using the cached chain is actually slower by a couple seconds in the test. I had based this off mining basic originally.

EyeOfPython: > You can't mine without a second node.

Really? It always worked for me, it's working on my regtest with just 1 node

EyeOfPython (approved):

PR #186, opened 2021-05-23
Allow minerfunding to be disabled on mainet/testnet

schancel:
Currently, minerfund is only allowed to be enabled on regtest. This
commit allows it do be disabled on mainnet and testnet, as well as in
unit tests.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -284,7 +284,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity) {
     // Note that by default, these tests run with size accounting enabled.
     GlobalConfig config;
     // Don't check this functionality in this test.
-    config.SetEnableMinerFund(false);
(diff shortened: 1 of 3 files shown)


src/test/miner_tests.cpp
```diff
@@ -284,7 +284,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity) {
     // Note that by default, these tests run with size accounting enabled.
     GlobalConfig config;
     // Don't check this functionality in this test.
-    config.SetEnableMinerFund(false);
+    config.SetEnableMinerFund(true);
```
EyeOfPython, line 287: What's this?

schancel: Setting it to false didn't do anything before, now it does so it needs to be turned back on.

EyeOfPython (approved):

PR #187, opened 2021-05-23
Remove checking of miner funding from CreateNewBlock_validity test

schancel:
This commit stops CreateNewBlock_validity from checking miner funding
in the coinbase, as well as makes finding txIds in the blocks more
generic. The second change makes updating the test less fragile. The
specific ordering is no longer important since the introduction of
CTOR.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -65,28 +65,28 @@ BlockAssembler MinerTestingSetup::AssemblerForTest(const Config &config) {
 }
 
 constexpr uint64_t blockinfo[] = {
-    57137586545,  203906838680, 37341269554,  124675956107, 143498055326,
-    34221506364,  104198109,    39235918446,  86470147682,  178412698046,
-    11529010165,  73535899488,  49116792746,  4431721905,   9850823019,
-    80454294224,  4848789644,   111086237654, 82682535872,  29915475741,
-    12095352580,  32556579783,  10415626579,  55545333028,  250331715143,
-    143491684625, 208187975679, 229394364557, 19522049784,  4465616070,
-    105968788127, 15209329897,  62225562945,  164025189649, 115661654471,
-    2917315162,   200928144,    34310199100,  126198469056, 209410321226,
-    40250710375,  51755687307,  37235251287,  133614149050, 4495244887,
-    52624989304,  37820597512,  822816206,    32349072859,  39871789317,
-    38787280109,  118391407386, 82189705523,  201514131224, 41242290287,
-    77234567619,  87283561757,  22476264135,  26251475603,  14932055915,
-    64240910770,  62018122820,  42432154561,  87143303142,  39627815544,
-    97667634757,  60373793386,  4253033578,   1538871139,   24950353753,
-    198216687206, 109331536301, 20430148983,  55232661289,  5211076681,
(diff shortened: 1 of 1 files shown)


src/test/miner_tests.cpp
```diff
@@ -776,6 +788,52 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity) {
...
+BOOST_AUTO_TEST_CASE(CreateNewBlock_minerfund) {
+    // Note that by default, these tests run with size accounting enabled.
+    GlobalConfig config;
+    // Don't check this functionality in this test.
+    config.SetEnableMinerFund(true);
+    const CChainParams &chainparams = config.GetChainParams();
+    CScript scriptPubKey =
+        CScript() << ParseHex("04678afdb0fe5548271967f1a67130b7105cd6a828e03909"
+                              "a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112"
+                              "de5c384df7ba0b8d578a4c702b6bf11d5f")
```
EyeOfPython, line 801: Can you use a different script? Uncompressed pubkey seems a bit unnecessary

src/test/miner_tests.cpp
```diff
@@ -776,6 +788,52 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity) {
...
+                  << OP_CHECKSIG;
+    std::unique_ptr<CBlockTemplate> pblocktemplate;
+    CMutableTransaction tx;
+    CScript script;
+    TestMemPoolEntryHelper entry;
+    entry.nFee = 11 * SATOSHI;
+    entry.nHeight = 11;
```
EyeOfPython, line 808: Doesn't seem to be used? Please check if you can simplify this test.

src/test/miner_tests.cpp
```diff
@@ -242,7 +247,14 @@ void MinerTestingSetup::TestPackageSelection(
     tx.vout[0].nValue = (100000000 - 10000) * SATOSHI;
     m_node.mempool->addUnchecked(entry.Fee(10000 * SATOSHI).FromTx(tx));
     pblocktemplate = AssemblerForTest(config).CreateNewBlock(scriptPubKey);
-    BOOST_CHECK_EQUAL(pblocktemplate->block.vtx[3]->GetId(), lowFeeTxId2);
+    const auto foundTx = std::find_if(
+        pblocktemplate->block.vtx.begin(), pblocktemplate->block.vtx.end(),
+        [lowFeeTxId2](const CTransactionRef &a) -> bool {
+            return a->GetId() == lowFeeTxId2;
+        });
```
EyeOfPython, line 254: Good. Thanks for cleaning this up finally

EyeOfPython (changes requested):

src/test/miner_tests.cpp
schancel: Yeah, I'll look through. Probably missed some copypasta.

src/test/miner_tests.cpp
schancel: This was surprisingly painful to figure out. Lack of defined operators on some of the Bitcoind types made my life difficult.

EyeOfPython (approved):
Looks good now.

PR #199, opened 2021-06-19
Fix lotus seeder checkpoints height

ngocviet:
Hi @schancel,

 I try to implement the fix for the seeder but I got some issues:
* When running the seeder on a vps, it can only crawl the directly connected node, not the other nodes.
* Even though the nslookup returns the node's addresses (I can see it while debugging), but no data is sync for lotusd (with config: dnsseed=1, connect=1)

My DNS config for the domain abcpay.cash is:

| Type | Host | Value |
|----------|----------|---|
| A Record | seeder | 167.179.88.165 |
| NS Record | dnsseed | seeder.abcpay.cash |

The command I ran the dnsseed is:
```
 sudo ./lotus-seeder -ns=seeder.abcpay.cash -mbox=test@gmail.com -chain=test -host=dnsseed.abcpay.cash -port=53
```

Could you help to check?

Thanks!

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -260,6 +260,8 @@ public:
         vFixedSeeds.clear();
         vSeeds.clear();
 
+        vSeeds.emplace_back("seeder.abcpay.cash");
+
         base58Prefixes[PUBKEY_ADDRESS] = std::vector<uint8_t>(1, 111);
         base58Prefixes[SCRIPT_ADDRESS] = std::vector<uint8_t>(1, 196);
         base58Prefixes[SECRET_KEY] = std::vector<uint8_t>(1, 239);
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -260,8 +260,6 @@ public:
         vFixedSeeds.clear();
         vSeeds.clear();
 
-        vSeeds.emplace_back("seeder.abcpay.cash");
-
         base58Prefixes[PUBKEY_ADDRESS] = std::vector<uint8_t>(1, 111);
         base58Prefixes[SCRIPT_ADDRESS] = std::vector<uint8_t>(1, 196);
         base58Prefixes[SECRET_KEY] = std::vector<uint8_t>(1, 239);
--- a/src/seeder/bitcoin.cpp
+++ b/src/seeder/bitcoin.cpp
@@ -61,8 +61,10 @@ PeerMessagingState CSeederNode::ProcessMessage(std::string strCommand,
         // nVersion);
         if (vAddr) {
             MessageWriter::WriteMessage(vSend, NetMsgType::GETADDR);
-            std::vector<BlockHash> locatorHash(
-                1, Params().Checkpoints().mapCheckpoints.rbegin()->second);
+            std::vector<BlockHash> locatorHash(0);
+            if (Params().Checkpoints().mapCheckpoints.size() >= 2) {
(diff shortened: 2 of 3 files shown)


schancel:
`sudo ./lotus-seeder -ns=seeder.abcpay.cash`

This should be `dnsseed.abcpay.cash` right?

src/chainparams.cpp
```diff
@@ -260,6 +260,8 @@ class CTestNetParams : public CChainParams {
         vFixedSeeds.clear();
         vSeeds.clear();
 
+        vSeeds.emplace_back("seeder.abcpay.cash");
```
schancel, line 263: I think this should be `dnsseed.abcpay.cash` I believe.

schancel:
Yeah you have the ns records specified backwards.

ngocviet:
Hi @schancel ,
I tried both your suggestion on change the vSeeds value or swap the -ns and -host value in seeder command but it still does not work for now.

I followed the guide at https://github.com/LogosFoundation/lotusd/tree/master/src/seeder
"seeder.abcpay.cash" equals to "vps.example.com" and "dnsseed.abcpay.cash" equals to "dnsseed.example.com" in the guide.
```cpp
vSeeds.emplace_back("seeder.abcpay.cash"); 
```
The domain "seeder.abcpay.cash" are used for 2 things:
* As the bootstrap address for the lotus-seeder's crawler. The seeder start to crawl from this domain, So "seeder.abcpay.cash" should be a A record points to a machine with a running node.
* As the seed address which is used by nodes to query all the associated IPs which the seeder return (if set seeder=1). But in the code the query address is "x1.seeder.abcpay.cash". If I use the option -overridednsseed=seeder.abcpay.cash then I can see the address returned successfully.

The prefix x1 is from the line:
```
std::string host = strprintf("x%x.%s", requiredServiceBits, seed);
```

I still think that the config is correct since if I use -overridednsseed option I can see the address return from the seeder, just wondering what is "x1" prefix and how should I configure it.


Hope that provides you more context.

src/seeder/bitcoin.cpp
```diff
@@ -61,8 +61,10 @@ PeerMessagingState CSeederNode::ProcessMessage(std::string strCommand,
         // nVersion);
         if (vAddr) {
             MessageWriter::WriteMessage(vSend, NetMsgType::GETADDR);
-            std::vector<BlockHash> locatorHash(
-                1, Params().Checkpoints().mapCheckpoints.rbegin()->second);
+            std::vector<BlockHash> locatorHash(0);
+            if (Params().Checkpoints().mapCheckpoints.size() >= 2) {
```
schancel, line 65: Don't understand why this should be >= 2 instead of 1?

EyeOfPython:
Here's what @KarolTrzeszczkowski told me:

> The [seeder] we've inherited from core is broken. Port some commits from BCHN or Bitcoin Static. I'll send you which ones in the morning, but it should be easy to find them in my repo, there is an accumulation of seeder related commits

He then linked commits https://github.com/Ergon-moe/Bitcoin-Static/commit/f29c7d82d4adcf2bea31bce0f4aabc76a18f4c04 to https://github.com/Ergon-moe/Bitcoin-Static/commit/a3cbe05ec348704f7d61f09602fc02cc917e271a on the Ergon node as a potential fix.

> They are mostly from sipa seeder, the main seeder repo
> https://github.com/sipa/bitcoin-seeder

schancel:
Seems to not be needed now?

PR #200, opened 2021-06-19
Merge v0.23.6

schancel:


src/amount.cpp
```diff
@@ -6,21 +6,34 @@
...
+#include <currencyunit.h>
 #include <network.h>
 #include <univalue.h>
-
+#include <util/system.h>
 #include <tinyformat.h>
 
+static const Currency CURRENCY_BCHA{COIN, SATOSHI, 8, "BCHA"};
+static const Currency CURRENCY_XEC{100 * SATOSHI, SATOSHI, 2, "XEC"};
+static const Currency CURRENCY_LOTUS{1000000 * SATOSHI, SATOSHI, 6, "Lotus"};
```
schancel, line 17: Should be XPI. I forgot.

EyeOfPython (approved):
On a later date, we should remove the -ecash flag, especially once it becomes on by default

PR #203, opened 2021-06-20
Update Exodus upgrade activation time

schancel:
Updating activation time to winter solstice 2021:
  2021-12-21T15:59:00.000Z

Obtained timing from: https://www.weather.gov/media/ind/seasons.pdf


EyeOfPython (approved):
I get
Unix Timestamp | 1640102340
-- | --
GMT | Tue Dec 21 2021 15:59:00 GMT+0000

Also checked on https://www.timeanddate.com/calendar/winter-solstice.html, which gives the same date.

So that looks good to me.

PR #204, opened 2021-06-20
Update base seeds for static persistent nodes

schancel:
Several people have setup persistent nodes online. This commit adds
them as we don't quite yet have DNS seeders up.


EyeOfPython (approved):
Nice

PR #215, opened 2021-08-20
New PoW-based epoch hash

EyeOfPython:
The previous implementation of the epoch hash would simply roll over the epoch every 5040 blocks.
In practice, this has very limited application, as each epoch block only covers 1 block worth of PoW.

With this change, a new epoch starts when the previous block hash is 5040 times below the current target.

In essence, this turns "very lucky" blocks into epoch blocks, which, on average, should occur every 5040 blocks.

"Faking" these blocks requires actually expending 5040 times the amount of PoW than of normal blocks.
This means that smart contracts and low-power devices can verify 5040 blocks worth of PoW by checking a single hash.

This vastly simplifies e.g. wrapping coins from other chains on Lotus.


src/init.cpp
```diff
@@ -410,7 +410,7 @@ void SetupServerArgs(NodeContext &node) {
         "-min", "-resetguisettings", "-rootcertificates=<file>", "-splash",
         "-uiplatform",
         // TODO remove after the November 2020 upgrade
-        "-axionactivationtime"};
+        "-axionactivationtime", "-exodusactivationtime"};
```
schancel, line 413: Should probably rename this so we ensure we get merge conflicts on any activation changes of theirs. That's what I did with the actual variable.

src/pow/pow.cpp
```diff
@@ -66,3 +66,28 @@ bool CheckProofOfWork(const BlockHash &hash, uint32_t nBits,
...
+    arith_uint256 bnTarget;
+
+    bnTarget.SetCompact(nBits, &fNegative, &fOverflow);
+    bnTarget /= EPOCH_NUM_BLOCKS;
+
+    // Check if proof of work is sufficient to be an epoch hash
+    return UintToArith256(hash) <= bnTarget;
+}
+
+uint256 GetNextEpochBlockHash(const CBlockHeader *block,
```
schancel, line 82: Makes sense

src/init.cpp
schancel: I guess what I'm saying is to remove `"-axionactivationtime"` and add this as a separate PR.

src/consensus/activation.cpp
```diff
@@ -4,4 +4,16 @@
...
+#include <util/system.h>
+
+bool IsExodusEnabled(const Consensus::Params &params,
+                     const CBlockIndex *pindexPrev) {
+    if (pindexPrev == nullptr) {
+        return false;
+    }
+
+    return pindexPrev->GetMedianTimePast() >=
+           gArgs.GetArg("-exodusactivationtime", params.exodusActivationTime);
```
schancel, line 18: Yeah, I thought I had added the activation time already.

schancel (changes requested):
Requesting changes to prevent accidental merge.

PR #223, opened 2021-09-18
Add NNG interface

EyeOfPython:
This allows external indexers to be built efficiently using an easy to use NNG interface


src/validation.cpp
```diff
@@ -739,7 +739,14 @@ bool MemPoolAccept::AcceptSingleTransaction(const CTransactionRef &ptx,
         return false;
     }
 
-    GetMainSignals().TransactionAddedToMempool(ptx);
+    std::vector<Coin> spent_coins;
+    spent_coins.reserve(ptx->vin.size());
```
schancel, line 743: Why do you need `spent_coins` here? Avoid having to look them up via more calls?

test/functional/interface_nng.py
```diff
@@ -0,0 +1,688 @@
+#!/usr/bin/env python3
```
schancel, line 1: I think the naming of this will trigger a testing error? There's some prefix checks.

schancel:
There is a `depends` folder as well as the gitian descriptors that are required. I *believe* that based on what you said that `flatbuffers` would need to have a makefile in `depends/packages` and the gitian builds need `libnng-dev`

src/validation.cpp
EyeOfPython: Yes, spent coins are not indexed. The data is right there in the node, so this simplifies the indexer.

schancel (approved):

test/functional/interface_nng.py
EyeOfPython: No, this obeys the naming rules.

PR #230, opened 2021-11-02
Refactor minerfund, remove random selection

EyeOfPython:
Previously, the coinbase outputs for the minerfund were randomly picked from a set of addresses based on `hashEpochBlock`.

However, establishing such a set turned out to be difficult. Therefore, the address sets always contained only one entry, so that address would always be picked.

This PR removes picking the addresses at random, in order to simplify adding activation code down the line.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -91,11 +91,9 @@ public:
 
         // The miner fund is enabled by default on mainnet.
         consensus.enableMinerFund = ENABLE_MINER_FUND;
-        consensus.payoutAddressSets = RewardAddresses::MainNet::AddressSets;
-        assert(consensus.payoutAddressSets.size() == 13);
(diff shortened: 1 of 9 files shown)


src/chainparams.cpp
```diff
@@ -218,11 +215,8 @@ class CTestNetParams : public CChainParams {
 
         // The miner fund is enabled by default on testnet.
         consensus.enableMinerFund = ENABLE_MINER_FUND;
-        consensus.payoutAddressSets = RewardAddresses::TestNet::AddressSets;
```
schancel, line 221: Would prefer this wasn't renamed. Miner fund is kind of presuming miners are paying these addresses. They are not. But I won't block the PR over it. Just don't want future social drama based on misunderstandings.

schancel (approved):

PR #231, opened 2021-11-02
Add `IsExodusEnabled`

EyeOfPython:
Checks whether the Exodus upgrade has activated yet, based on MTP.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -418,8 +418,8 @@ void SetupServerArgs(NodeContext &node) {
         "-allowselfsignedrootcertificates", "-choosedatadir", "-lang=<lang>",
         "-min", "-resetguisettings", "-rootcertificates=<file>", "-splash",
         "-uiplatform",
-        // TODO remove after the November 2020 upgrade
-        "-axionactivationtime"};
+        // TODO remove after the December 2021 upgrade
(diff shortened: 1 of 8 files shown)


src/init.cpp
```diff
@@ -418,7 +418,7 @@ void SetupServerArgs(NodeContext &node) {
         "-min", "-resetguisettings", "-rootcertificates=<file>", "-splash",
         "-uiplatform",
         // TODO remove after the November 2020 upgrade
-        "-axionactivationtime"};
+        "-axionactivationtime", "-exodusactivationtime"};
```
schancel, line 421: This other flag isn't being used except by the test suite right?

EyeOfPython: Yes, seems like it.

schancel (approved):
Looks good. We will need to change the locations of `axionactivation` to have this though so the functional tests run the activation code.

src/init.cpp
schancel: Rather than adding this flag, it might be good to find & replace all axionactivation times. There's stuff in a variety of files that cause the test on (for example, this diff) to run tests with the activated changes.

schancel (approved):

PR #232, opened 2021-11-02
Update coinbase addresses for Exodus hardfork

EyeOfPython:
Coinbase addresses are going to update during the Exodus hardfork to new recipients.


schancel (changes requested):
Looks great. Thank you. Let's merge once we fix the addresses and also we should change the other PR to properly activate exodus w/ the functional tests.

schancel (commented):
Looks good. Just need to update the addresses now.

schancel (approved):

PR #234, opened 2021-11-12
Add `leviticusactivationtime`, bump replay protecc

EyeOfPython:
Nodes automatically fork off from the network if not upgraded at the determined timestamp.
This change bumps the "fork off" timestamp (replayprotectiontime) to 2022-06-21 09:14:00, such that the node continues to accept normally signed (non-replay protected) transactions at the Exodus network upgrade.


schancel (approved):
Gracias!

PR #239, opened 2022-01-04
Use BASEPREFIX in platform toolchain cmake files

EyeOfPython:
Currently, the platform toolchain cmake files (Linux32.cmake, OSX.cmake, etc.) all use CMAKE_CURRENT_SOURCE_DIR to build the toolchain path.
This breaks however once the toolchain files are used from a different source directory, e.g. when building dependencies.
By threading the BASEPREFIX through the build system, the toolchain files can also be used by dependencies.


shammahatdivvy:
This looks like it might need to be ported up to ABC?

schancel (approved):

PR #242, opened 2022-01-04
Add `spent_coins` to TransactionAddedToMempool

EyeOfPython:
Access to `spent_coins` is required for the NNG interface. This adds it to `CValidationInterface`.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -739,7 +739,15 @@ bool MemPoolAccept::AcceptSingleTransaction(const CTransactionRef &ptx,
         return false;
     }
 
-    GetMainSignals().TransactionAddedToMempool(ptx);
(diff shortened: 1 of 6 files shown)


schancel (approved):

src/validation.cpp
```diff
@@ -739,7 +739,15 @@ bool MemPoolAccept::AcceptSingleTransaction(const CTransactionRef &ptx,
         return false;
     }
 
-    GetMainSignals().TransactionAddedToMempool(ptx);
+    std::vector<Coin> spent_coins;
+    spent_coins.reserve(ptx->vin.size());
```
schancel, line 743: This seems to be done a lot, might be worth creating a function on the transaction classes.

PR #247, opened 2022-02-26
Add epoch merkle root metadata field

EyeOfPython:


schancel (commented):
May be worthwhile to split the epoch hash changes from this metadata change?

src/consensus/extendedmetadata.cpp
```diff
@@ -0,0 +1,153 @@
...
+    epochMerkleLeaves.push_back(
+        ComputeMerkleRoot({uint256(), block.hashMerkleRoot}, dummyNumLayers));
+
+    // Walk back chain until epoch hash changes
+    while (pindexPrev && pindexPrev->hashEpochBlock == currentEpochBlockHash) {
+        // Add H(block hash || merkle root)
+        epochMerkleLeaves.push_back(ComputeMerkleRoot(
+            {pindexPrev->GetBlockHash(), pindexPrev->hashMerkleRoot},
+            dummyNumLayers));
+        pindexPrev = pindexPrev->pprev;
```
schancel, line 37: Do we need some check in here. I realize we have so many blocks as to not hit the genesis after this is activated, but seems prudent?

src/consensus/extendedmetadata.cpp
```diff
@@ -0,0 +1,153 @@
...
+std::vector<uint256> CollectEpochMerkleLeaves(const CBlock &block,
+                                              const CBlockIndex *pindexPrev) {
+    const uint256 currentEpochBlockHash = block.hashEpochBlock;
+    std::vector<uint256> epochMerkleLeaves;
+    size_t dummyNumLayers;
+
+    // Add block's own merkle root
+    // 0 for the block hash, as we cannot know our own block hash here
+    epochMerkleLeaves.push_back(
+        ComputeMerkleRoot({uint256(), block.hashMerkleRoot}, dummyNumLayers));
```
schancel, line 29: I don't understand why we would want this.

schancel (changes requested):
Requesting changes to ensure no accidental merge.

PR #272, opened 2022-09-04
Merge Bitcoin-ABC v0.24.4 into Lotusd

PiRK:
Note that `ninja check-secp256k1` tests and feature_pruning.py (`ninja check-extended`) are failing, but failures are not related to this backport, they also happen on master.


PiRK:
Some functional tests still failing.

test/functional/wallet_send.py
```diff
@@ -0,0 +1,424 @@
...
+        self.nodes[0].generate(1)
+        assert_equal(self.nodes[0].gettransaction(txid)["confirmations"], 1)
+
+        self.log.info("Lock unspents...")
+        utxo1 = w0.listunspent()[0]
+        assert_greater_than(utxo1["amount"], 1)
+        res = self.test_send(
+            from_wallet=w0,
+            to_wallet=w1,
+            amount=100,
```
schancel, line 400: The assert here doesn't end up matching the amount sent.

schancel:
Tests don't fail locally. Seems these are flakey tests?

schancel:
Pretty sure the nng_interface test is flaky due to signature size happening to be 72 instead of 73 for that run. Probably should fix in a separate PR though.

PiRK:
Fixed the failing interface_nng test by updating a data position after pad_tx changed (tx is now 100 bytes rather than 101).

The wallet_send test failure is intermittent and is already fixed in Bitcoin ABC. It will be fixed in a future merge. It is basically a missing `sync_all` that causes some recently mined coins to sometimes not be available.

schancel:
Manually merged to master 🦾

PR #274, opened 2022-09-10
[test] Use transaction sizes for checking block position

schancel:
The current nng_interface test uses hardcoded locations in the block data. This causes the test to break any time the transaction sizes change due to changes in other areas of the code which generate transactions. (e.g. a differently sized ecdsa transaction, or pad_tx being updated)


PiRK (approved):
Screenshot from 2022-09-11 10-08-54

PR #275, opened 2022-09-16
Merge Bitcoin-ABC v0.24.5 into Lotusd

PiRK:
That one was easy.


PiRK:
Need to investigate failing ZMQ tests.

PiRK:
The hardcoded amount of 200 XPI to spend in  `create_conflicting_tx` was too large in some situations (when the UTXO to spend was not a full block reward, but rather the change of a previous transaction), so the calculation  `utxo["amount"] - Decimal("200.1")` could in some situations be a negative amount.

Fixed by changing the amount to a lower value of 5 XPI. This should not break unless `create_conflicting_tx` is called more than 40 times in between two blocks in the future.

schancel:
Master updated to branch. Thanks much!

PR #277, opened 2022-10-24
Add numbersactivationtime and, bump replay protecc time

schancel:
Nodes automatically fork off from the network if not upgraded at the
determined timestamp. This change bumps the "fork off" timestamp
(replayprotectiontime) to 2023-06-21T14:58:00.000Z (Deuteronomy
upgrade), such that the node continues to accept normally signed
(non-replay protected) transactions at the "Numbers" network upgrade.

### DIFF
--- a/src/consensus/activation.cpp
+++ b/src/consensus/activation.cpp
@@ -27,3 +27,18 @@ bool IsLeviticusEnabled(const Consensus::Params &params,
     return pindexPrev->GetMedianTimePast() >=
            gArgs.GetArg("-leviticusactivationtime", params.leviticusActivationTime);
 }
+
+bool IsNumbersEnabled(const Consensus::Params &params,
+                      const int64_t nMedianTimePast) {
+    return nMedianTimePast >=
+           gArgs.GetArg("-numbersactivationtime", params.numbersActivationTime);
+}
+
+bool IsNumbersEnabled(const Consensus::Params &params,
+                      const CBlockIndex *pindexPrev) {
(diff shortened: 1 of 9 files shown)


schancel:
Following checklist (https://github.com/LogosFoundation/lotusd/wiki/Upgrade-Checklist)

- [ ] Update way coinbase funds are handled.
- [ ] Re-add standardness checks

schancel:
- [ ] Need to add isnumbersenabled test.

src/consensus/activation.cpp
```diff
@@ -27,3 +27,18 @@ bool IsLeviticusEnabled(const Consensus::Params &params,
     return pindexPrev->GetMedianTimePast() >=
            gArgs.GetArg("-leviticusactivationtime", params.leviticusActivationTime);
 }
+
+bool IsNumbersEnabled(const Consensus::Params &params,
```
schancel, line 31: adding this as some locations only have access to the height, and not the pindex.

EyeOfPython: where is this needed?

If you're 100% sure it's needed, also add some tests

EyeOfPython (changes requested):

src/consensus/activation.cpp
schancel: There is a test. You can look at ABC code for now it's needed. But GetNextBlockScriptFlags() is where it's used.

EyeOfPython (approved):

PR #286, opened 2022-11-06
Add exodus checkpoint + fix leviticus checkpoint

schancel:
Add in the exodus checkpoint for recordkeeping, and also fix the leviticus checkpoint as it was one block after the upgrade. The miner fund addresses change one block after each upgrade.

### DIFF
--- a/src/networks/abc/checkpoints.cpp
+++ b/src/networks/abc/checkpoints.cpp
@@ -7,9 +7,13 @@
 
 static CCheckpointData mainNetCheckpointData = {
     .mapCheckpoints = {
+        // Exodus activation
+        {143509, BlockHash::fromHex("00000000000df2347fd05d10e9ce5846f3089ef7e6"
+                                    "0b650a2725f9c462c5b601")},
         // Leviticus activation
-        {273562, BlockHash::fromHex("0000000000049e558f98d76240a0469720dee51bc2"
-                                    "7fd4c7f4e8946b6c50fb99")},
+        {273561, BlockHash::fromHex("000000000006a214caf1b62fdc5fa9746aa90d2887"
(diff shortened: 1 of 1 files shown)


src/networks/abc/checkpoints.cpp
```diff
@@ -7,9 +7,13 @@
 
 static CCheckpointData mainNetCheckpointData = {
     .mapCheckpoints = {
+        // Exodus activation
+        {143509, BlockHash::fromHex("00000000000df2347fd05d10e9ce5846f3089ef7e6"
```
EyeOfPython, line 11: Checkpoint the block with the upgraded rules, which is one later

EyeOfPython (changes requested):

src/networks/abc/checkpoints.cpp
schancel: Technically not true. Only the funding addresses update one later. Any other rules will activate at this block.

EyeOfPython: This is the code for the funding addresses:

```

static const std::vector<std::string> &
GetPayoutAddresses(const Consensus::Params &params,
                   const CBlockIndex *pindexPrev) {
    if (IsLeviticusEnabled(params, pindexPrev)) {
        return params.coinbasePayoutAddresses.leviticus;
    }
    if (IsExodusEnabled(params, pindexPrev)) {
        return params.coinbasePayoutAddresses.exodus;
    }
    return params.coinbasePayoutAddresses.genesis;
}
```

`IsExodusEnabled` and `IsLeviticusEnabled` will be used exactly the same for all other rule updates, no?

EyeOfPython (approved):

PR #287, opened 2022-11-06
Fix numbers activation time to be winter solstice

schancel:
Due to dyslexia, the replace activation time was set to `2022-12-12T21:48:00.000Z` instead of `2022-12-21T21:48:00.000Z`. This will impact miners and users insomuch as they will mine empty blocks -- but only if they do not upgrade. As everyone will need to upgrade in order to get the new functionality for the numbers upgrade, this will be fine.


maff1989 (approved):
Looks good

EyeOfPython (approved):

Issue #280, opened 2022-11-05
Disallow Taproot in numbers upgrade


schancel:
As per title. Let's disallow this going forward so we don't have to maintain an unusued feature.

PR #289, opened 2022-11-07
Phase out Taproot and SIGHASH_LOTUS

EyeOfPython:
1. Reject transactions with Taproot outputs from being added to the mempool. We do this by adding TxHasPayToTaproot, which is checked in `MemPoolAccept::PreChecks`.
2. Add `SCRIPT_DISABLE_TAPROOT_SIGHASH_LOTUS` script flag, which makes the interpreter return an error when the user wants to spend a taproot script, or use the SIGHASH_LOTUS sighash type.
3. Add  `SCRIPT_DISABLE_TAPROOT_SIGHASH_LOTUS` to `STANDARD_SCRIPT_VERIFY_FLAGS`.
4. Before Numbers activates, remove `SCRIPT_DISABLE_TAPROOT_SIGHASH_LOTUS` from required script flags so we don't split the network.
5. After Numbers activated, transactions with Taproot output are now violating consensus.


EyeOfPython:
Closes #280.

schancel (approved):

PR #294, opened 2022-11-14
Fix `update-chainparams.sh` so it can be ran on MacOS X

schancel:
Script now looks for an env var called `CONFIG_FILE` that will override the default config location. Can be ran as:

```
CONFIG_FILE=CONFIG_FILE=~/Library/Application\ Support/Lotus/lotus.conf contrib/source-control-tools/patch-recipes/update-chainparams.sh
```


schancel:
I've verified the update, as I had to run this.

PR #295, opened 2022-11-14
Update seeds lists

schancel:
This was manually updated in several places. The seeders are strangly not seeding a variety of nodes. There are only a few available. In addition to the nodes which the seeder is dumping, I have manually updated the list with peers from my own node, and the node which has been running the explorer.


maff1989 (commented):
FYI the Explorer node is at `173.208.206.181:10605`

maff1989 (approved):
Looks good

